PR #55, opened 2022-11-07
Sc 10279/tenant update

daniellemaxwell:
### Scope of changes

Implementation of Tenant Update API definition on the client-side.

Fixes SC-10279

### Type of change

- [x] new feature
- [ ] bug fix
- [ ] documentation
- [x] testing
- [ ] technical debt
- [ ] other (describe)

### Acceptance criteria

### Author checklist

### Reviewer(s) checklist


pkg/tenant/api/v1/client.go
```diff
@@ -153,6 +153,26 @@ func (s *APIv1) TenantDetail(ctx context.Context, id string) (out *Tenant, err e
 	return out, nil
 }
 
+func (s *APIv1) TenantUpdate(ctx context.Context, in *Tenant) (out *Tenant, err error) {
+	var id string
+	if in.ID != "" {
+		id = in.ID
+	}
+
+	path := fmt.Sprintf("/v1/tenant/%s", id)
```
bbengfort, line 162: We'll want a bit of error handling here; by default `id` is empty string so if no `in.ID` is supplied then it will be a `PUT /v1/tenant` which will return a 405 Method Not Allowed error -- which can be confusing to the user. 

```suggestion
	if in.ID != "" {
		return nil, errors.New("tenant id is required for update")
	}

	path := fmt.Sprintf("/v1/tenant/%s", in.ID)
```

bbengfort (approved):
Looks good! One small comment and it should be good to go!

pkg/tenant/api/v1/client.go
daniellemaxwell: Thank you for the explanation!

PR #57, opened 2022-11-10
Changes apps to projects and update api file

daniellemaxwell:
### Scope of changes

The following client methods were created along with corresponding tests:
`TenantProjectList`
`TenantProjectCreate`
`ProjectList`
`ProjectCreate`

To set params, the size (PageSize) and tok (NextPageToken) arguments were added to each list method.

Fixes SC-10349

### Type of change

- [x] new feature
- [ ] bug fix
- [ ] documentation
- [x] testing
- [ ] technical debt
- [ ] other (describe)

### Acceptance criteria

### Author checklist

### Reviewer(s) checklist


pkg/tenant/api/v1/client.go
```diff
@@ -100,9 +100,9 @@ func (s *APIv1) SignUp(ctx context.Context, in *ContactInfo) (err error) {
 	return nil
 }
 
-func (s *APIv1) TenantList(ctx context.Context, in PageQuery) (out *TenantPage, err error) {
-	// Set values for PageSize and NextPageToken keys in the PageQuery struct
-	in = PageQuery{2, "12"}
+func (s *APIv1) TenantList(ctx context.Context, size uint32, tok string, in PageQuery) (out *TenantPage, err error) {
+	// Set values for PageSize and NextPageToken in PageQuery
+	in = PageQuery{size, tok}
```
daniellemaxwell, line 105: Here's a link to the query package to I used to set the values for PageSize and NextPageToken: https://pkg.go.dev/github.com/google/go-querystring@v1.1.0/query#Values

bbengfort: I think that this package is a good choice for binding struct values into query params! 

There does seem to be a bit of duplication here; you're having the user pass in `in PageQuery` but then you're overwriting what the user passes in with the `size` and `tok` elements. Wouldn't it be better to just have the user pass in the PageQuery without constructing it again, e.g.

```
func (s *APIv1) TenantList(ctx context.Context, in PageQuery) (out *TenantPage, err error) {
```

Then, if we change the PageQuery parameter (e.g. if we add `PrevPageToken` for backwards pagination) then the API method signatures would not have to change.

pkg/tenant/api/v1/client.go
```diff
@@ -288,6 +289,105 @@ func (s *APIv1) MemberCreate(ctx context.Context, in *Member) (out *Member, err
 	return out, nil
 }
 
+func (s *APIv1) TenantProjectList(ctx context.Context, id string, size uint32, tok string, in PageQuery) (out *TenantProjectPage, err error) {
+	if id == "" {
+		return nil, ErrMemberIDRequired
```
bbengfort, line 294: Should this be `ErrProjectIDRequired`?

pkg/tenant/api/v1/client.go
```diff
@@ -288,6 +289,105 @@ func (s *APIv1) MemberCreate(ctx context.Context, in *Member) (out *Member, err
...
+	out = &TenantProjectPage{}
+	if _, err = s.Do(req, out, true); err != nil {
+		return nil, err
+	}
+	return out, nil
+}
+
+func (s *APIv1) TenantProjectCreate(ctx context.Context, id string, in *Project) (out *Project, err error) {
+	if id == "" {
+		return nil, ErrMemberIDRequired
```
bbengfort, line 322: Same as above, `ErrProjectIDRequired`?

bbengfort (commented):
Progress!

pkg/tenant/api/v1/client.go
daniellemaxwell: You're right. I need add an error for each resource type.

pkg/tenant/api/v1/client.go
daniellemaxwell: I looked at the bff pkg and figured out why I was having trouble with this. I've made the updates and will push shortly.

pkg/tenant/api/v1/client.go
daniellemaxwell: This will also be ErrTenantIDRequired and then ErrProjectIDRequired will be used for the remaining definitions.

pkg/tenant/api/v1/client.go
```diff
@@ -247,24 +248,23 @@ func (s *APIv1) TenantMemberCreate(ctx context.Context, id string, in *Member) (
...
+func (s *APIv1) MemberList(ctx context.Context, in *PageQuery) (out *MemberPage, err error) {
 	var params url.Values
 	if params, err = query.Values(in); err != nil {
 		return nil, fmt.Errorf("could not encode query params: %w", err)
 	}
+
 	// Make the HTTP request
 	var req *http.Request
-	if req, err = s.NewRequest(ctx, http.MethodGet, "v1/members", nil, &params); err != nil {
+	if req, err = s.NewRequest(ctx, http.MethodGet, "/v1/members", nil, &params); err != nil {
```
bbengfort, line 259: Good catch!

bbengfort (approved):
Looks good, thanks for making those changes!

PR #60, opened 2022-11-16
APIKey create and list client methods

daniellemaxwell:
### Scope of changes

Added the following client methods and corresponding tests:
`ProjectAPIKeyList`
`ProjectAPIKeyCreate`
`APIKeyList`
`APIKeyCreate`

This is a short review that should take a few minutes.

Fixes SC-10359

### Type of change

- [x] new feature
- [ ] bug fix
- [ ] documentation
- [x] testing
- [ ] technical debt
- [ ] other (describe)

### Acceptance criteria

### Author checklist

### Reviewer(s) checklist


pkg/tenant/api/v1/api.go
```diff
@@ -128,6 +134,29 @@ type TopicPage struct {
...
 
+type ProjectAPIKeyPage struct {
+	ProjectID     string `json:"project_id"`
+	APIKeys       []*APIKey
+	PrevPageToken string
+	NextPageToken string
+}
+
+type APIKey struct {
+	ID          int      `json:"id,omitempty"`
```
daniellemaxwell, line 145: I noticed in quarterdeck there's a `ClientID`. I thought that the `TenantID` would be the equivalent of that here, so didn't include one. However, if you think `TenantID` or a `ClientID` should be included, let me know.

bbengfort: The `ClientID` is like a username and it's a component of the API Key - so it's a little different than the database ID. 

The way that the API Key requests work is that you can create an API key with Name and ProjectID and then you'll get back the ClientID and ClientSecret - but this is the _only_ time that the ClientSecret will be visible in the application or to the user (after that it's gone forever). Users can update the name of the API Key, and delete it, but that's it. 

We may want to take another pass at this when the quarterdeck API is complete.

bbengfort (approved):
Thanks for updating all of those `context.TODO()` elements!

pkg/tenant/api/v1/api.go
daniellemaxwell: I appreciate the explanation. I've added `ClientID` and changed `KeySecret` to `ClientSecret`.

PR #69, opened 2022-11-22
TenantCreate handler

daniellemaxwell:
### Scope of changes

- Adds TestCreateTenant server-side handler along with corresponding test. 
- Replaces JSON with msgpack for marshaling and unmarshaling in the tenants db.
- Replaces uuid with ulid for sorting IDs.
- Blank test files for the remaining resources have been added as well for minor organization.

Fixes SC-10272

### Type of change

- [x] new feature
- [ ] bug fix
- [ ] documentation
- [x] testing
- [ ] technical debt
- [ ] other (describe)

### Acceptance criteria

### Author checklist

- [x]  I have updated the dependencies list

### Reviewer(s) checklist


pkg/tenant/db/tenants.go
```diff
@@ -30,18 +31,16 @@ func (t *Tenant) Namespace() string {
...
 func (t *Tenant) UnmarshalValue(data []byte) error {
-	// TODO: look into bson, msgpack, etc.
-	return json.Unmarshal(data, t)
+	return msgpack.Unmarshal(data, t)
 }
 
 func CreateTenant(ctx context.Context, tenant *Tenant) (err error) {
-	if tenant.ID == uuid.Nil {
-		tenant.ID = uuid.New()
+	if tenant.ID.String() == "" {
```
daniellemaxwell, line 42: ulid.Nil isn't provided by the ulid package, so I chose to go this route. However, I could change the code and follow the example here: https://github.com/oklog/ulid/issues/84#issuecomment-1254728179

pkg/tenant/db/tenants.go
```diff
@@ -30,18 +31,16 @@ func (t *Tenant) Namespace() string {
...
+		tenant.ID = ulid.Make()
```
daniellemaxwell, line 43: I used Make() for now, but am thinking I should probably replace with New() as mentioned in the ReadMe here: https://github.com/oklog/ulid

What do you think?

pkg/tenant/db/tenants.go
```diff
@@ -2,19 +2,20 @@ package db
...
 type Tenant struct {
-	ID       uuid.UUID
-	Name     string
-	Created  time.Time
-	Modified time.Time
+	ID              ulid.ULID `msgpack:"id"`
+	Name            string    `msgpack:"name"`
+	EnvironmentType string    `msgpack:"environment_type"`
+	Created         time.Time `msgpack:"created"`
+	Modified        time.Time `msgpack:"modified"`
```
bbengfort, line 18: Thank you for adding the field names!

pkg/tenant/db/tenants.go
bbengfort: Yes, you're probably right -- let's create a story for this. I think we could create a small `ulid` package that wraps the `oklog/ulid` package and has the `IsZero()` functionality mentioned above and also has a `Make` function that uses cryptographic security and monotonic entropy. We'll also probably have to ensure thread safety.

bbengfort, line 43:
```suggestion
		// TODO: use crypto rand and monotonic entropy with ulid.New
		tenant.ID = ulid.Make()
```

pkg/tenant/db/tenants.go
```diff
@@ -67,7 +66,7 @@ func RetrieveTenant(ctx context.Context, id uuid.UUID) (tenant *Tenant, err erro
 }
 
 func UpdateTenant(ctx context.Context, tenant *Tenant) (err error) {
-	if tenant.ID == uuid.Nil {
+	if tenant.ID.String() == "" {
```
bbengfort, line 69: Same comparison as above.

pkg/tenant/db/tenants_test.go
```diff
@@ -59,58 +59,65 @@ func (s *dbTestSuite) TestCreateTenant() {
 	require.NoError(err, "could not create tenant")
 
 	// Fields should have been populated
-	require.NotEqual(uuid.Nil, tenant.ID, "expected non-zero uuid to be populated")
+	require.NotEqual("", tenant.ID, "expected non-zero ulid to be populated")
```
bbengfort, line 62: I suspect that `NotZero` should work with a ulid -- but I'm not 100% sure. 

```suggestion
	require.NotZero(tenant.ID, "expected non-zero ulid to be populated")
```

pkg/tenant/db/tenants_test.go
```diff
@@ -59,58 +59,65 @@ func (s *dbTestSuite) TestCreateTenant() {
...
-			return nil, status.Errorf(codes.FailedPrecondition, "could not read fixture: %s", err)
-		}
+		// Marshal the data with msgpack
+		data, err := tenant.MarshalValue()
+		require.NoError(err, "could not marshal the tenant")
+
+		// Unmarshal the data with msgpack
+		other := &db.Tenant{}
+		err = other.UnmarshalValue(data)
+		require.NoError(err, "could not unmarshal the tenant")
```
bbengfort, line 94: Should we create a follow on story for msgpack fixture helpers? RE: the golden files blog post I sent you yesterday?

pkg/tenant/tenants.go
```diff
@@ -0,0 +1,121 @@
...
+
+	if err = c.BindJSON(&tenant); err != nil {
+		log.Warn().Err(err).Msg("could not parse tenant create request")
+		c.JSON(http.StatusBadRequest, api.ErrorResponse("could not parse request"))
+		return
+	}
+
+	if tenant.ID == "" {
+		c.JSON(http.StatusBadRequest, api.ErrorResponse(api.ErrTenantIDRequired))
+	}
```
bbengfort, line 28: I think this should be the opposite - when we create a tenant, we don't want the user to specify an ID - we want to assign the ID ourselves. 

```suggestion
	if tenant.ID != "" {
		c.JSON(http.StatusBadRequest, api.ErrorResponse(api.ErrInvalidTenantField))
		return
	}
```

pkg/tenant/tenants.go
```diff
@@ -0,0 +1,121 @@
...
+
+	// Add Tenant to the database
+	if err = db.CreateTenant(c.Request.Context(), &db.Tenant{ID: ulid.Make()}); err != nil {
+		log.Error().Err(err).Msg("could not save tenant")
+		c.JSON(http.StatusInternalServerError, api.ErrorResponse("could not add tenant"))
+		return
+	}
+
+	c.JSON(http.StatusCreated, tenant)
```
bbengfort, line 45: Per the comment above, we'll likely want to ensure that we update the response with the ID, created, and modified timestamps that get generated during the `db.CreateTenant` process.

pkg/tenant/tenants.go
```diff
@@ -0,0 +1,121 @@
...
+
+	"github.com/gin-gonic/gin"
+	"github.com/oklog/ulid/v2"
+	"github.com/rotationalio/ensign/pkg/tenant/api/v1"
+	"github.com/rotationalio/ensign/pkg/tenant/db"
+	"github.com/rs/zerolog/log"
+)
+
+// CreateTenant creates a new tenant in the database
+func (s *Server) CreateTenant(c *gin.Context) {
```
bbengfort, line 14: We will still need some information from authentication here; the organization and user -- I'm working hard to get that finalized as we speak!

pkg/tenant/tenants.go
```diff
@@ -0,0 +1,121 @@
...
+
+	c.JSON(http.StatusCreated, tenant)
+}
+
+func (s *Server) TenantList(c *gin.Context) {
+
+	c.JSON(http.StatusNotImplemented, "not implemented yet")
+}
+
+func (s *Server) TenantCreate(c *gin.Context) {
```
bbengfort, line 53: See above comments :-) 

Could we also add docstrings above the method names to describe what they do and why `CreateTenant` and `TenantCreate` are different methods?

pkg/tenant/tenants.go
```diff
@@ -0,0 +1,121 @@
...
+		out *api.Reply
+	)
+
+	tenantID := c.Param("tenantID")
+
+	if tenantID == "" {
+		log.Error().Err(err).Msg(db.ErrMissingID.Error())
+		c.JSON(http.StatusInternalServerError, api.ErrorResponse("could not retrieve tenant"))
+		return
+	}
```
bbengfort, line 106: You're pretty much guaranteed to get some string as the `tenantID` otherwise this handler would not be called. However, if the tenantID is not a ULID, we should return a 404 since the parameter is part of the url. E.g. if the request is to `DELETE /v1/tenants/foo` -- `foo` is a 404 since it's not a ULID. 

```suggestion
	tenantID := c.Param("tenantID")
	var tenantID ulid.ULID
	if tenantID, err = ulid.Parse(c.Param("tenantID")); err != nil {
		log.Debug().Err(err).Msg("could not parse tenant ulid")
		c.JSON(http.StatusNotFound, api.ErrorResponse(ErrTenantNotFound))
		return
	}
```

pkg/tenant/tenants.go
```diff
@@ -0,0 +1,121 @@
...
+
+	// TODO: Replace uuid with ulid
+	req := &db.Tenant{
+		ID: ulid.Make(),
+	}
```
bbengfort, line 111: `req` is not being used so we can simply delete these lines. The request is coming from the param that is parsed above. 

```suggestion
```

pkg/tenant/tenants.go
```diff
@@ -0,0 +1,121 @@
...
+
+	// Delete Tenant and
+	if err = db.DeleteTenant(c.Request.Context(), req.ID); err != nil {
```
bbengfort, line 114: We should delete the ID that is parsed from the url param - not a random ULID generated by `Make()`. 

```suggestion
	if err = db.DeleteTenant(c.Request.Context(), tenantID); err != nil {
```

pkg/tenant/tenants.go
```diff
@@ -0,0 +1,121 @@
...
+		log.Error().Err(err).Msg(db.ErrMissingID.Error())
```
bbengfort, line 115: We should probably talk about systematizing error logging -- I actually haven't figured this out yet, but it's probably important. For now what I'll say is to keep in mind that any `log.Error()` message will go to the slack channel and should cause us to take action. E.g. we're hoping that if there is a `Delete` error it means something has gone horribly wrong. We want to make sure we give as much information as possible in the log message:

```suggestion
		log.Error().Err(err).Str("tenantID", tenantID.String()).Msg("could not delete tenant")
```

pkg/tenant/tenants.go
```diff
@@ -0,0 +1,121 @@
...
+		c.JSON(http.StatusInternalServerError, api.ErrorResponse("could not retrieve tenant"))
```
bbengfort, line 116:
```suggestion
		c.JSON(http.StatusInternalServerError, api.ErrorResponse("could not delete tenant"))
```

pkg/tenant/tenants.go
```diff
@@ -0,0 +1,121 @@
...
+func (s *Server) TenantCreate(c *gin.Context) {
+	var (
+		err    error
+		tenant *api.Tenant
+	)
+
+	// TODO: Add authentication and authorization middleware
+	if err = c.BindJSON(&tenant); err != nil {
+		log.Warn().Err(err).Msg("could not parse tenant create request")
+		c.JSON(http.StatusBadRequest, api.ErrorResponse("could not bind request"))
```
bbengfort, line 62: Don't forget to return after calling `c.JSON`! 

```suggestion
		c.JSON(http.StatusBadRequest, api.ErrorResponse("could not bind request"))
		return
```

pkg/tenant/tenants.go
```diff
@@ -0,0 +1,121 @@
...
+	}
+
+	if tenant.EnvironmentType == "" {
+		log.Error().Err(err).Msg("tenant does not have an environment type")
+	}
+
+	// TODO: Add tenant to the database
+	if err = db.CreateTenant(c.Request.Context(), &db.Tenant{ID: ulid.Make()}); err != nil {
+		log.Error().Err(err).Msg("could not save tenant")
+		c.JSON(http.StatusInternalServerError, api.ErrorResponse("could not add tenant"))
```
bbengfort, line 80:
```suggestion
		c.JSON(http.StatusInternalServerError, api.ErrorResponse("could not add tenant"))
		return
```

pkg/tenant/tenants.go
```diff
@@ -0,0 +1,121 @@
...
+	if tenant.Name == "" {
+		log.Warn().Err(err).Msg("tenant does not have an name")
+	}
+
+	if tenant.EnvironmentType == "" {
+		log.Error().Err(err).Msg("tenant does not have an environment type")
+	}
+
+	// TODO: Add tenant to the database
+	if err = db.CreateTenant(c.Request.Context(), &db.Tenant{ID: ulid.Make()}); err != nil {
```
bbengfort, line 78: We don't want to create a random tenant, we want to create the tenant that the user passed in:

```suggestion
	if err = db.CreateTenant(c.Request.Context(), tenant); err != nil {
```

pkg/tenant/tenants.go
```diff
@@ -0,0 +1,121 @@
...
+		log.Error().Err(err).Msg("tenant does not have an ID")
+	}
+
+	if tenant.Name == "" {
+		log.Warn().Err(err).Msg("tenant does not have an name")
+	}
+
+	if tenant.EnvironmentType == "" {
+		log.Error().Err(err).Msg("tenant does not have an environment type")
+	}
```
bbengfort, line 75: We don't need `log` messages on bad inputs since the user can fix the problem and send the request again (log messages are for us to debug problems) - unless you want to issue `log.Debug()` so we can figure out what the user is doing wrong. 

Don't forget to return an `http.StatusBadRequest` here and stop the handler with `return`!

pkg/tenant/tenants.go
```diff
@@ -0,0 +1,121 @@
...
+	if tenant.Name == "" {
+		c.JSON(http.StatusBadRequest, api.ErrorResponse("tenant name is required"))
+	}
+
+	if tenant.EnvironmentType == "" {
+		c.JSON(http.StatusBadRequest, api.ErrorResponse("environment type is required"))
+	}
+
+	// Add Tenant to the database
+	if err = db.CreateTenant(c.Request.Context(), &db.Tenant{ID: ulid.Make()}); err != nil {
```
bbengfort, line 39: We should create the tenant that the user passed in :-) 

```suggestion
	if err = db.CreateTenant(c.Request.Context(), tenant); err != nil {
```

pkg/tenant/tenants.go
```diff
@@ -0,0 +1,121 @@
...
+		c.JSON(http.StatusBadRequest, api.ErrorResponse(api.ErrTenantIDRequired))
+	}
+
+	if tenant.Name == "" {
+		c.JSON(http.StatusBadRequest, api.ErrorResponse("tenant name is required"))
+	}
+
+	if tenant.EnvironmentType == "" {
+		c.JSON(http.StatusBadRequest, api.ErrorResponse("environment type is required"))
+	}
```
bbengfort, line 36: Don't forget to return after sending bad request (I always make this mistake too):

```suggestion
	if tenant.Name == "" {
		c.JSON(http.StatusBadRequest, api.ErrorResponse("tenant name is required"))
		return
	}

	if tenant.EnvironmentType == "" {
		c.JSON(http.StatusBadRequest, api.ErrorResponse("environment type is required"))
		return
	}
```

bbengfort (commented):
Great start!

pkg/tenant/db/tenants_test.go
daniellemaxwell: Yes, it'll work. I'll have to add an ID to that test.

pkg/tenant/db/tenants_test.go
daniellemaxwell: Yes, I'm going to create a story about that today.

pkg/tenant/tenants.go
daniellemaxwell: While dealing with an error (the one related to getting the mock) I decided to test a delete method to see if I'd get the same issue. I thought I'd deleted this in my last commit. Oh well. I'll save this for when I review the detail and delete changes before submitting a PR.

pkg/tenant/tenants.go
daniellemaxwell: Thank you for providing more details about this.

pkg/tenant/tenants_test.go
```diff
@@ -0,0 +1,38 @@
...
+	require.Error(err, http.StatusBadRequest, "expected unimplemented error")
+
+	req := &db.Tenant{
+		Name:            "tenant01",
+		EnvironmentType: "prod",
+	}
+
+	tenant, err := suite.client.TenantCreate(ctx, req)
+	require.NoError(err, "could not add tenant")
+	require.NotEmpty(tenant.ID, "tenant id should not be zero")
```
daniellemaxwell, line 35: I chose `NotEmpty` because the db will generate an id, but maybe `NotZero` would be better.

pkg/tenant/db/tenants.go
```diff
@@ -67,8 +67,9 @@ func RetrieveTenant(ctx context.Context, id uuid.UUID) (tenant *Tenant, err erro
 }
 
 func UpdateTenant(ctx context.Context, tenant *Tenant) (err error) {
-	if tenant.ID == uuid.Nil {
-		return ErrMissingID
+	if tenant.ID.Compare(ulid.ULID{}) == 0 {
+		// TODO: use crypto rand and monotonic entropy with ulid.New
+		tenant.ID = ulid.Make()
```
bbengfort, line 72: I think we still want to return `ErrMissingID` here on update; otherwise we will be creating a new Tenant record. 

```suggestion
		return ErrMissingID
```

pkg/tenant/db/tenants.go
```diff
@@ -30,18 +31,17 @@ func (t *Tenant) Namespace() string {
...
 func (t *Tenant) UnmarshalValue(data []byte) error {
-	// TODO: look into bson, msgpack, etc.
-	return json.Unmarshal(data, t)
+	return msgpack.Unmarshal(data, t)
 }
 
 func CreateTenant(ctx context.Context, tenant *Tenant) (err error) {
-	if tenant.ID == uuid.Nil {
-		tenant.ID = uuid.New()
+	if tenant.ID.Compare(ulid.ULID{}) == 0 {
```
bbengfort, line 42: When we add docstrings to these methods (and we should) we should note that if the User passes in an ID then it will be used; otherwise a new ID will be created.

pkg/tenant/tenants.go
```diff
@@ -0,0 +1,76 @@
...
+		c.JSON(http.StatusBadRequest, api.ErrorResponse(api.ErrTenantIDRequired))
+	}
+
+	if t.Name == "" {
+		c.JSON(http.StatusBadRequest, api.ErrorResponse("tenant name is required"))
+	}
+
+	if t.EnvironmentType == "" {
+		c.JSON(http.StatusBadRequest, api.ErrorResponse("environment type is required"))
+	}
```
bbengfort, line 38: Don't forget to `return` after an error! 

```suggestion
	if t.ID != "" {
		c.JSON(http.StatusBadRequest, api.ErrorResponse(api.ErrTenantIDRequired))
		return
	}

	if t.Name == "" {
		c.JSON(http.StatusBadRequest, api.ErrorResponse("tenant name is required"))
		return
	}

	if t.EnvironmentType == "" {
		c.JSON(http.StatusBadRequest, api.ErrorResponse("environment type is required"))
		return
	}
```

pkg/tenant/tenants.go
```diff
@@ -0,0 +1,76 @@
...
+
+	// TODO: Add authentication and authorization middleware
+
+	if err = c.BindJSON(&t); err != nil {
+		log.Warn().Err(err).Msg("could not bind tenant create request")
+		c.JSON(http.StatusBadRequest, api.ErrorResponse("could not bind request"))
+	}
+
+	if t.ID != "" {
+		c.JSON(http.StatusBadRequest, api.ErrorResponse(api.ErrTenantIDRequired))
```
bbengfort, line 29: Shouldn't this error be something like "cannot specify tenant ID on create" rather than "tenant ID required"?

pkg/tenant/tenants.go
```diff
@@ -0,0 +1,76 @@
...
+		err error
+		t   *api.Tenant
+	)
+
+	// TODO: Add authentication and authorization middleware
+
+	if err = c.BindJSON(&t); err != nil {
+		log.Warn().Err(err).Msg("could not bind tenant create request")
+		c.JSON(http.StatusBadRequest, api.ErrorResponse("could not bind request"))
+	}
```
bbengfort, line 26:
```suggestion
	if err = c.BindJSON(&t); err != nil {
		log.Warn().Err(err).Msg("could not bind tenant create request")
		c.JSON(http.StatusBadRequest, api.ErrorResponse("could not bind request"))
		return
	}
```

pkg/tenant/tenants.go
```diff
@@ -0,0 +1,76 @@
...
+		c.JSON(http.StatusBadRequest, api.ErrorResponse("tenant name is required"))
+	}
+
+	if t.EnvironmentType == "" {
+		c.JSON(http.StatusBadRequest, api.ErrorResponse("environment type is required"))
+	}
+
+	tenantID := ulid.Make()
+	created := time.Now()
+	modified := created
```
bbengfort, line 42: Is it necessary to do this work here? Won't the `db` package do this?

pkg/tenant/tenants.go
```diff
@@ -0,0 +1,76 @@
...
+		EnvironmentType: t.EnvironmentType,
+		Created:         created,
+		Modified:        modified,
+	}
+
+	// Add tenant to the database
+	if err = db.CreateTenant(c.Request.Context(), tenant); err != nil {
+		log.Error().Err(err).Msg("could not create tenant in database")
+		c.JSON(http.StatusInternalServerError, api.ErrorResponse("could not add tenant"))
+	}
```
bbengfort, line 56:
```suggestion
	if err = db.CreateTenant(c.Request.Context(), tenant); err != nil {
		log.Error().Err(err).Msg("could not create tenant in database")
		c.JSON(http.StatusInternalServerError, api.ErrorResponse("could not add tenant"))
		return
	}
```

bbengfort (approved):
Looks good! I had a couple of minor comments but we can think about them in the next PR if needed.

pkg/tenant/tenants.go
daniellemaxwell: I'll update the error to be more accurate. After all making changes to find the error, I didn't properly update this.

pkg/tenant/tenants.go
daniellemaxwell: Thanks! I had those in and then tried going back to a previous commit while trying to solve an error and mostly out of frustration didn't remember to make this change.

pkg/tenant/tenants.go
daniellemaxwell: I got this mixed up with a previous comment about updating the response with what's generated from the db and will make the change.

PR #73, opened 2022-12-09
Sc 10284/update handler

daniellemaxwell:
### Scope of changes

Adds TenantUpdate server-side handler that will update tenant records in the Tenant API along with corresponding tests.

Fixes SC-10284

### Type of change

- [x] new feature
- [ ] bug fix
- [ ] documentation
- [x] testing
- [ ] technical debt
- [ ] other (describe)

### Acceptance criteria

### Author checklist

### Reviewer(s) checklist


pkg/tenant/tenants.go
```diff
@@ -139,6 +138,62 @@ func (s *Server) TenantDelete(c *gin.Context) {
...
+	// Bind the user request with JSON and return a 400 response if
+	// binding is not successful.
+	if err = c.BindJSON(&tenant); err != nil {
+		log.Warn().Err(err).Msg("could not parse tenant update request")
+		c.JSON(http.StatusBadRequest, api.ErrorResponse("could not bind request"))
+		return
+	}
+
+	// Get the tenant ID from the URL and return a 400 if the tenant
+	// does not exist.
```
pdeziel, line 162:
```suggestion
	// ID is not a ULID
```

pkg/tenant/tenants.go
```diff
@@ -139,6 +138,62 @@ func (s *Server) TenantDelete(c *gin.Context) {
...
+		c.JSON(http.StatusBadRequest, api.ErrorResponse("tenant environment type is required"))
+		return
+	}
+
+	// Prepare update request for insertion into the database.
+	req := &db.Tenant{
+		ID:              tenantID,
+		Name:            tenant.Name,
+		EnvironmentType: tenant.EnvironmentType,
+	}
```
pdeziel, line 188: The pattern we've mostly used in GDS for Update handlers is to first try to retrieve the original record, then update the record that was retrieved and store it back to the database.

I think this would go a long way to make this endpoint more resilient to future changes, since it prevents us from accidentally overwriting internal data on the tenant in the database. We may also have metadata fields like `created_at`, `modified_at` timestamps that we don't want to override. This would also allow us to handle the "tenant not found" and "error calling the database" errors separately to give the caller more information about what went wrong.

pkg/tenant/tenants_test.go
```diff
@@ -85,6 +85,42 @@ func (suite *tenantTestSuite) TestTenantDetail() {
...
+	// Should return an error if the tenant name does not exist
+	_, err := suite.client.TenantUpdate(ctx, &api.Tenant{ID: "01ARZ3NDEKTSV4RRFFQ69G5FAV", EnvironmentType: "prod"})
+	suite.requireError(err, http.StatusBadRequest, "tenant name is required", "expected error when tenant name does not exist")
+
+	// Should return an error if the tenant environment type does not exist
+	_, err = suite.client.TenantUpdate(ctx, &api.Tenant{ID: "01ARZ3NDEKTSV4RRFFQ69G5FAV", Name: "example-dev"})
+	suite.requireError(err, http.StatusBadRequest, "tenant environment type is required", "expected error when tenant environent type does not exist")
+
+	req := &api.Tenant{
+		ID:              "01ARZ3NDEKTSV4RRFFQ69G5FAV",
```
pdeziel, line 112: Can we change this to a different ID so that we're testing that the ID in the request does not override the ID on the tenant object?

pdeziel (commented):
I had one suggestion to make sure we're not overwriting data on the tenant objects in the database but other than that it's looking good! Thanks for continuing to add the tests as we create the endpoints!

pkg/tenant/tenants.go
daniellemaxwell: @pdeziel I made an update to retrieve the tenant from the db once the tenant ID is parsed from the URL. However, if the change isn't in line with what you had in mind please let me know.

pdeziel: Great, now that we are retrieving the tenant from the db can we use that instead of the `req` (so modifying the model directly) when storing to the database?

Also, I think it's probably better to put the `db.RetrieveTenant` after the 400 checks, so that we don't have to make a request to the database unless the request is formatted correctly (this will save a bit of time on the error paths).

pkg/tenant/tenants.go
```diff
@@ -138,6 +138,71 @@ func (s *Server) TenantDelete(c *gin.Context) {
...
+		ID:              t.ID,
+		Name:            tenant.Name,
+		EnvironmentType: tenant.EnvironmentType,
+	}
+
+	// Update tenant in the database and return a 404 response if the
+	// tenant record cannot be updated.
+	if err := db.UpdateTenant(c.Request.Context(), req); err != nil {
+		log.Error().Err(err).Msg("could not save tenant")
+		c.JSON(http.StatusNotFound, api.ErrorResponse("could not update tenant"))
```
pdeziel, line 203:
```suggestion
		c.JSON(http.StatusInternalServerError, api.ErrorResponse("could not update tenant"))
```

pdeziel (approved):
Thanks, I had a few more suggestions to shore up the endpoint but after that looks good!

pkg/tenant/tenants.go
pdeziel: At this point we have already verified that the tenant exists so if we get an error here then it's probably a server error.

PR #74, opened 2022-12-12
Sc 11487/list tenants

daniellemaxwell:
### Scope of changes

This will add `ListTenants` to the tenants db package. This addition to the db will allow retrieve a list of tenants assigned to a workspace. If approved, this allow for the implementation of the `TenantList` server-side handler that uses the client.

List takes in a context, prefix, and namespace. For reference, the db's `List` function may be viewed [here](https://github.com/rotationalio/ensign/blob/main/pkg/tenant/db/db.go) and the test file is [here](https://github.com/rotationalio/ensign/blob/main/pkg/tenant/db/db_test.go).

Fixes SC-11487

### Type of change

- [x] new feature
- [ ] bug fix
- [ ] documentation
- [x] testing
- [ ] technical debt
- [ ] other (describe)

### Acceptance criteria

### Author checklist

### Reviewer(s) checklist


pkg/tenant/db/tenants_test.go
```diff
@@ -66,6 +67,35 @@ func (s *dbTestSuite) TestCreateTenant() {
...
 }
 
+func (s *dbTestSuite) TestListTenants() {
+	require := s.Require()
+	ctx := context.Background()
+
+	prefix := []byte("test")
+	namespace := "testing"
+
+	s.mock.OnCursor = func(in *pb.CursorRequest, stream pb.Trtl_CursorServer) error {
```
daniellemaxwell, line 77: Following a discussion with @bbengfort it was decided to use `Cursor` for now.

pkg/tenant/db/tenants.go
```diff
@@ -55,6 +55,15 @@ func CreateTenant(ctx context.Context, tenant *Tenant) (err error) {
...
 }
 
+// ListTenants retrieves all tenants assigned to an organization.
+func ListTenants(ctx context.Context, prefix []byte, namespace string) (values [][]byte, err error) {
+
+	if values, err = List(ctx, prefix, namespace); err != nil {
+		return nil, err
+	}
+	return values, err
+}
```
bbengfort, line 65: I'm so glad we get to start to discuss the data model for tenants even further now! 

The spec for `ListTenants` as described by the doc string says that we want to get all tenants for an organization. We already have a constant for our namespace, `TenantNamespace` on line 11, so I recommend that we change the signature of this method so that we take as input an `orgID` and we return a slice of `*Tenant` already parsed from the data. 

```suggestion
// ListTenants retrieves all tenants assigned to an organization.
func ListTenants(ctx context.Context, orgID ulid.ULID) (tenants []*Tenant, err error) {
	// TODO: ensure that the tenants are stored with the orgID as their prefix! 
	var prefix []byte
	
	// TODO: it would be better to use the cursor directly rather than have duplicate data in memory
	var values [][]byte
	if values, err = List(ctx, prefix, TenantNamespace); err != nil {
		return nil, err
	}
	
	// Parse the tenants from the data
	tenants = make([]*Tenant, 0, len(values)) 
	for _, data := range values {
		tenant := &Tenant{}
		if err = tenant.UnmarshalValue(data); err != nil {
			return nil, err
		}
		tenants = append(tenants, tenant) 
	}
	
	return tenants, nil
}
```

There are a couple of TODOs there which I will describe below.

**TODO 1**: In order to fetch all of the tenants by organization, the organization ID needs to be a prefix of the key. The trtl database stores keys in lexicographical order. So consider the "tenants" namespace (e.g. the dataset of all tenants); it's keys might look something like:

tenant001 (org001)
tenant002 (org002)
tenant003 (org003)
tenant004 (org001)
tenant005 (org003)

When we list values, we're always going to scan from the top of the list to the bottom. The database has very fast seek (skip values) and partial iteration, however, which is why it's so fast to Get tenant003. 

A prefix is basically a seek to the first item with the specific prefix and then continue iterating until the next key doesn't have the prefix. If we change our keys to instead be:

org001-tenant001
org001-tenant004
org002-tenant002
org003-tenant003
org003-tenant005

Then we could use the prefix "org003" to return tenant003 and tenant005. 

*Note* - if the prefix is nil then all tenants are returned. 

So we need to do a couple things to make this happen.

1. We need to add OrgID to the `Tenant` model
2. We need to update the `Tenant.Key()` method to include the `OrgID` as a prefix
3. We need to update the `List` method to convert the `orgID` into a `[]byte` prefix. 

**BUT WAIT! There is another option!** (sorry worst PR review ever) 

The prefix based method of grouping related tenants is the _fastest_ way to retrieve tenants. But we can only do it with one type of prefix, so we want to ensure that the most list common query uses the prefix method (and that's definitely OrgID for tenants … so we're going to do what I wrote above … just wanted to make sure you understood all the options). The alternative is to use an index to look up all of the tenant keys belonging to the OrgID. There are two ways to do this:

1. Store the IDs of the tenants that belong to the org on an `Organization` model 
2. Create a "mapping" collection that is orgID -> list of tenant IDs. 

You would then do a Get for the index key then use the value to Get each individual organization. We're going to have to do this with some of our other models, which is why I wanted to bring it up here.

**TODO 2**: We're creating two slices in memory, the values `[][]byte` and the tenants array `[]*Tenant` it would be better if we used the Cursor directly and only had one array in memory … but we can create a follow on story for this optimization later and leave the TODO in the code.

bbengfort (commented):
Let me know if you want to chat about what I wrote below!

pkg/tenant/db/tenants.go
daniellemaxwell: I was thinking that an orgID would need to be added. Thank you for the explanation and also for mentioning the alternative.

bbengfort (approved):
Looks great - nice work on the new key structure!

PR #77, opened 2022-12-19
MemberDetail/Delete handlers, tests, and update comments

daniellemaxwell:
### Scope of changes

Adds the MemberDetail and MemberDelete server-side handlers to the Tenant API. MemberDetail will provide a summary of a member by a given ID and MemberDelete will delete the record of a member by a given ID.

Fixes SC-10408

### Type of change

- [x] new feature
- [ ] bug fix
- [ ] documentation
- [x] testing
- [ ] technical debt
- [ ] other (describe)

### Acceptance criteria

### Author checklist

### Reviewer(s) checklist

### DIFF
--- a/pkg/tenant/members.go
+++ b/pkg/tenant/members.go
@@ -4,6 +4,10 @@ import (
 	"net/http"
 
 	"github.com/gin-gonic/gin"
+	"github.com/oklog/ulid/v2"
+	"github.com/rotationalio/ensign/pkg/tenant/api/v1"
+	"github.com/rotationalio/ensign/pkg/tenant/db"
+	"github.com/rs/zerolog/log"
 )
 
 func (s *Server) TenantMemberList(c *gin.Context) {
@@ -31,9 +35,40 @@ func (s *Server) MemberList(c *gin.Context) {
 	c.JSON(http.StatusNotImplemented, "not implemented yet")
 }
 
+// MemberDetail retrieves a summary detail of a member by its ID
+// and returns a 200 OK response.
+//
+// Route: /member/:memberID
 func (s *Server) MemberDetail(c *gin.Context) {
+	var (
+		err   error
+		reply *api.Member
+	)
 
-	c.JSON(http.StatusNotImplemented, "not implemented yet")
+	// Get the member ID from the URL and return a 400 if the member does not exist.
+	var memberID ulid.ULID
+	if memberID, err = ulid.Parse(c.Param("memberID")); err != nil {
+		log.Debug().Err(err).Msg("could not parse member ulid")
+		c.JSON(http.StatusBadRequest, api.ErrorResponse("could not parse member id"))
+		return
+	}
+
+	// Get the specified member from the database and return a 404 response
+	// if it cannot be retrieved.
+	var member *db.Member
+	if member, err = db.RetrieveMember(c.Request.Context(), memberID); err != nil {
+		log.Error().Err(err).Msg("could not retrieve member")
+		c.JSON(http.StatusNotFound, api.ErrorResponse("could not retrieve member"))
+		return
+	}
+
+	reply = &api.Member{
+		ID:   member.ID.String(),
+		Name: member.Name,
+		Role: member.Role,
+	}
+
+	c.JSON(http.StatusOK, reply)
 }
 
 func (s *Server) MemberCreate(c *gin.Context) {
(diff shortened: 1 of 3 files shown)


pkg/tenant/tenants.go
```diff
@@ -94,7 +94,7 @@ func (s *Server) TenantDetail(c *gin.Context) {
 		return
 	}
 
-	// Get the specified tenant from the database and return a 500 response
+	// Get the specified tenant from the database and return a 404 response
```
daniellemaxwell, line 97: Comments were updated to match the correct status codes that will be returned.

pkg/tenant/members.go
```diff
@@ -31,9 +35,40 @@ func (s *Server) MemberList(c *gin.Context) {
...
+		log.Debug().Err(err).Msg("could not parse member ulid")
+		c.JSON(http.StatusBadRequest, api.ErrorResponse("could not parse member id"))
+		return
+	}
+
+	// Get the specified member from the database and return a 404 response
+	// if it cannot be retrieved.
+	var member *db.Member
+	if member, err = db.RetrieveMember(c.Request.Context(), memberID); err != nil {
+		log.Error().Err(err).Msg("could not retrieve member")
```
pdeziel, line 60:
```suggestion
		log.Error().Err(err).Str("memberID", memberID.String()).Msg("could not retrieve member")
```

We might as well include this so we can tell which member ID the error is about.

pkg/tenant/members_test.go
```diff
@@ -1 +1,83 @@
...
+	defer trtl.Reset()
+
+	// Call the OnDelete method and return a DeleteReply.
+	trtl.OnDelete = func(ctx context.Context, dr *pb.DeleteRequest) (*pb.DeleteReply, error) {
+		return &pb.DeleteReply{}, nil
+	}
+
+	// Should return an error if the member does not exist.
+	err := suite.client.MemberDelete(ctx, "invalid")
+	suite.requireError(err, http.StatusBadRequest, "could not parse member id", "expected error when member does not exist")
```
pdeziel, line 79: Can we add a test case for when the member doesn't exist? In other words, it's parseable but the trtl delete returns an error.

pkg/tenant/tenants.go
pdeziel: Thanks!

pdeziel (approved):
I had a few minor suggestions and one suggestion for an additional test case but other than that looks great!

pkg/tenant/members.go
daniellemaxwell: I added this to TenantDetail as well.

pkg/tenant/members.go
```diff
@@ -31,9 +35,40 @@ func (s *Server) MemberList(c *gin.Context) {
...
+	var (
+		err   error
+		reply *api.Member
+	)
 
-	c.JSON(http.StatusNotImplemented, "not implemented yet")
+	// Get the member ID from the URL and return a 400 if the member does not exist.
+	var memberID ulid.ULID
+	if memberID, err = ulid.Parse(c.Param("memberID")); err != nil {
+		log.Debug().Err(err).Msg("could not parse member ulid")
```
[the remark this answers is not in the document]
daniellemaxwell: I mimicked what was approved for TenantDetail, but there wasn't a specific reason for why `Debug` was selected. I've updated this and MemberDetail to use `Error` for consistency.

pkg/tenant/members_test.go
daniellemaxwell: I added a test below, but I'm not sure if this is exactly what you meant.

pdeziel: I was thinking of something like this:

```go
trtl.OnDelete = func(ctx context.Context, dr *pb.DeleteRequest) (*pb.DeleteReply, error) {
    return nil, errors.New("key not found")
}

err = suite.client.MemberDelete(ctx, "missingID")
suite.RequireError(err, http.StatusNotFound, "could not delete member", "expected error when member ID is not found")
```

daniellemaxwell: Replacing `&pb.DeleteReply{}, nil` in the OnDelete function will result in the final error check returning an error. 

Also, when checking against `http.StatusNotFound` the following error message appears: 
`expected: "[404] could not delete member"`
`actual  : "[400] could not parse member id"`

pdeziel: Gotcha, does it work if you move the new OnDelete and the new test case to the end of the method? Also my bad on the example, the ID needs to be a real ulid since the handler needs to parse it first.

daniellemaxwell: It works now. Thank you!

PR #79, opened 2022-12-22
TenantMemberCreate and MemberCreate server-side handlers and tests

daniellemaxwell:
### Scope of changes

Adds server-side handlers for TenantMemberCreate and MemberCreate to the Tenant API. This will allow new members to be added to the database.

Fixes SC-10404

### Type of change

- [x] new feature
- [ ] bug fix
- [ ] documentation
- [x] testing
- [ ] technical debt
- [ ] other (describe)

### Acceptance criteria

### Author checklist

### Reviewer(s) checklist

### DIFF
--- a/pkg/tenant/members.go
+++ b/pkg/tenant/members.go
@@ -14,8 +14,57 @@ func (s *Server) TenantMemberList(c *gin.Context) {
 	c.JSON(http.StatusNotImplemented, "not implemented yet")
 }
 
+// / TenantMemberCreate adds a new tenant member to the database and returns
+// a 201 StatusCreated response.
+//
+// Route: /tenant/:tenantID/member
 func (s *Server) TenantMemberCreate(c *gin.Context) {
-	c.JSON(http.StatusNotImplemented, "not implemented yet")
+	var (
+		err    error
+		member *api.Member
+	)
+
+	// Bind the user request with JSON and return a 400 response if
+	// biding is not successful.
+	if err = c.BindJSON(&member); err != nil {
+		log.Warn().Err(err).Msg("could not bind tenant member create request")
+		c.JSON(http.StatusBadRequest, api.ErrorResponse("could not bind request"))
+		return
+	}
+
+	// Verify that a member ID does not exist and return a 400 response if
+	// the member id exists.
+	if member.ID != "" {
+		c.JSON(http.StatusBadRequest, api.ErrorResponse("member id cannot be specified on create"))
+		return
+	}
+
+	// Verify that a member name has been provided and return a 400 repsonse if
+	// the member name does not exist.
+	if member.Name == "" {
+		c.JSON(http.StatusBadRequest, api.ErrorResponse("tenant member name is required"))
+		return
+	}
+
+	// Verify that a member role has been provided and return a 400 response if
+	// the member role does not exist.
+	if member.Role == "" {
+		c.JSON(http.StatusBadRequest, api.ErrorResponse("tenant member role is required"))
+		return
+	}
+
+	tmember := &db.Member{
+		Name: member.Name,
+		Role: member.Role,
+	}
+
+	if err = db.CreateMember(c.Request.Context(), tmember); err != nil {
+		log.Error().Err(err).Msg("could not create tenant member in the database")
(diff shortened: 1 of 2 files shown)


pkg/tenant/members.go
```diff
@@ -14,8 +14,57 @@ func (s *Server) TenantMemberList(c *gin.Context) {
...
+		c.JSON(http.StatusBadRequest, api.ErrorResponse("could not bind request"))
+		return
+	}
+
+	// Verify that a member ID does not exist and return a 400 response if
+	// the member id exists.
+	if member.ID != "" {
+		c.JSON(http.StatusBadRequest, api.ErrorResponse("member id cannot be specified on create"))
+		return
+	}
```
bbengfort, line 40: The member has a `TenantID` field right? For this endpoint we should probably also parse the `:tenantID` from the url and make sure that is set on the `member`. 

```suggestion
	// Verify that a member ID does not exist and return a 400 response if
	// the member id exists.
	if member.ID != "" || member.TenantID != "" {
		c.JSON(http.StatusBadRequest, api.ErrorResponse("member and tenant id cannot be specified on create"))
		return
	}
	
	// TODO: parse :tenantID from URL and add to the member for create.
```

pkg/tenant/members.go
```diff
@@ -14,8 +14,57 @@ func (s *Server) TenantMemberList(c *gin.Context) {
...
+	// the member role does not exist.
+	if member.Role == "" {
+		c.JSON(http.StatusBadRequest, api.ErrorResponse("tenant member role is required"))
+		return
+	}
+
+	tmember := &db.Member{
+		Name: member.Name,
+		Role: member.Role,
+	}
```
bbengfort, line 59: Add member `TenantID` from the URL param.

pkg/tenant/members.go
```diff
@@ -35,6 +84,57 @@ func (s *Server) MemberList(c *gin.Context) {
...
+		c.JSON(http.StatusBadRequest, api.ErrorResponse("could not bind request"))
+		return
+	}
+
+	// Verify that a member id does not exist and return a 400 response if
+	// the member id exists.
+	if member.ID != "" {
+		c.JSON(http.StatusBadRequest, api.ErrorResponse("member is cannot be specified on create"))
+		return
+	}
```
bbengfort, line 110: In this case, we will want to assert that the `TenantID` is on the member struct and is parseable as a ULID right?

pkg/tenant/members.go
```diff
@@ -35,6 +84,57 @@ func (s *Server) MemberList(c *gin.Context) {
...
+
+	if member.Role == "" {
+		c.JSON(http.StatusBadRequest, api.ErrorResponse("member role is required"))
+		return
+	}
+
+	m := &db.Member{
+		Name: member.Name,
+		Role: member.Role,
+	}
```
bbengfort, line 127: The user will have to supply tenant ID here, but we need to parse it as a ULID into the `db.Member` struct.

bbengfort (approved):
Looks good - a few questions about the `tenantID` - but a good start to creating members! 

Note that we'll also have to have handling on these endpoints to send a request to Quarterdeck with the member information.

pkg/tenant/members.go
daniellemaxwell: There isn't a TenantID associated with the  api.Member struct, only an id for the member.  There's a TenantID on the TenantMemberPage struct that'll be used with TenantMemberList. This is why I was wondering if there should be separate member models in the db.

pkg/tenant/members.go
daniellemaxwell: The api.Member struct doesn't include TenantID. I can do c.Param("tenantID") to parse the ulid from the URL.

bbengfort: Oh, sorry if I'm forgetting, how are we listing members; by OrgID?

pkg/tenant/members.go
bbengfort: Ok, let's pause on this and follow up on it later - it might be Quarterdeck related. Go ahead and merge this PR.

pkg/tenant/members.go
daniellemaxwell: The api.Member struct currently only has a MemberID.

PR #80, opened 2022-12-23
Sc 11476/list handler

daniellemaxwell:
### Scope of changes

Adds the TenantList handler to the Tenant API. This will provide a list of user's assigned to a tenant by a given organization id.

Fixes SC-11476

### Type of change

- [x] new feature
- [ ] bug fix
- [ ] documentation
- [x] testing
- [ ] technical debt
- [ ] other (describe)

### Acceptance criteria

### Author checklist

### Reviewer(s) checklist

### DIFF
--- a/pkg/tenant/tenants.go
+++ b/pkg/tenant/tenants.go
@@ -12,22 +12,19 @@ import (
 
 // TenantList retrieves all tenants assigned to an organization and
 // returns a 200 OK response.
+//
+// Route: /tenant
 func (s *Server) TenantList(c *gin.Context) {
-	var (
-		err error
-		out *api.TenantPage
-	)
-
 	// Get tenants from the database and return a 500 response if not successful.
 	var tenant *db.Tenant
-	if _, err = db.ListTenants(c.Request.Context(), tenant.OrgID); err != nil {
+	if _, err := db.ListTenants(c.Request.Context(), tenant.OrgID); err != nil {
 		log.Error().Err(err).Msg("could not fetch tenants from database")
 		c.JSON(http.StatusInternalServerError, api.ErrorResponse("could not fetch tenants from the database"))
 		return
 	}
 
 	// Build the response
-	out = &api.TenantPage{Tenants: make([]*api.Tenant, 0)}
+	out := &api.TenantPage{Tenants: make([]*api.Tenant, 0)}
 
 	out.Tenants = append(out.Tenants, &api.Tenant{
 		ID:              tenant.ID.String(),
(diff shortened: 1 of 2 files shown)


pkg/tenant/tenants_test.go
```diff
@@ -11,6 +11,31 @@ import (
...
+	}
+
+	req := &api.PageQuery{
+		PageSize:      2,
+		NextPageToken: "12",
+	}
+
+	tenant, err := suite.client.TenantList(ctx, req)
+	require.NoError(err, "could not list tenants")
+	require.Len(tenant, 7)
```
daniellemaxwell, line 36: I had a test for this and meant to delete before creating the draft PR.

pkg/tenant/tenants_test.go
```diff
@@ -11,6 +11,31 @@ import (
 	"github.com/trisacrypto/directory/pkg/trtl/pb/v1"
 )
 
+func (suite *tenantTestSuite) TestTenantList() {
```
daniellemaxwell, line 14: I get the following message when trying to run the test: `runtime error: invalid memory address or nil pointer dereference`

The stack trace points to line 20 of the tenants.go file. I would think it's because I've used a `_`, but it doesn't seem like I can use a variable declared there. Although, I could be missing something.

pkg/tenant/tenants.go
```diff
@@ -10,8 +10,29 @@ import (
...
 
+// TenantList retrieves all tenants assigned to an organization and
+// returns a 200 OK response.
+//
+// Route: /tenant
 func (s *Server) TenantList(c *gin.Context) {
-	c.JSON(http.StatusNotImplemented, "not implemented yet")
+	// Get tenants from the database and return a 500 response if not successful.
+	var tenant *db.Tenant
+	if _, err := db.ListTenants(c.Request.Context(), tenant.OrgID); err != nil {
```
bbengfort, line 20: The problem is that `tenant` (declared on line 19) is `nil` so when you try to do `tenant.OrgID` - you're trying to get the property of a `nil` struct, which is causing the panic. 

The `orgID` will come from the JWT credentials of the user trying to list the tenants; so authentication is needed before you can implement the "list filtering". For now, I think you should pass in an empty ULID (e.g. `ulid.ULID{}`) and add a TODO that the `orgID` must be fetched. The TenantList will then return _all_ tenants in the database; but you can write TODOs in the test to expect that the list is filtered so only the tenants for the logged in user are returned.

pkg/tenant/tenants_test.go
bbengfort: See comment above on line 20

bbengfort (approved):
Nice one! Let's create a follow on story for filtering by organization ID.

PR #81, opened 2022-12-27
Sc 11781/topics db

daniellemaxwell:
### Scope of changes

Adds topics db package that will be used by the Tenant API's server-side handlers for the topics resource.

Fixes SC-11781

### Type of change

- [x] new feature
- [ ] bug fix
- [ ] documentation
- [x] testing
- [ ] technical debt
- [ ] other (describe)

### Acceptance criteria

### Author checklist

### Reviewer(s) checklist

### DIFF
--- a/pkg/tenant/db/projects_test.go
+++ b/pkg/tenant/db/projects_test.go
@@ -83,7 +83,7 @@ func (s *dbTestSuite) TestRetrieveProject() {
 
 	s.mock.OnGet = func(ctx context.Context, in *pb.GetRequest) (*pb.GetReply, error) {
 		if len(in.Key) == 0 || in.Namespace != db.ProjectNamespace {
-			return nil, status.Error(codes.FailedPrecondition, "bad Put request")
+			return nil, status.Error(codes.FailedPrecondition, "bad Get request")
 		}
 		if !bytes.Equal(in.Key[16:], project.ID[:]) {
 			return nil, status.Error(codes.NotFound, "project not found")
@@ -162,7 +162,7 @@ func (s *dbTestSuite) TestListProjects() {
 	projects = append(projects, project)
 	require.Len(projects, 1)
 
-	_, err = db.ListMembers(ctx, project.TenantID)
+	_, err = db.ListProjects(ctx, project.TenantID)
 	require.Error(err, "could not list projects")
 }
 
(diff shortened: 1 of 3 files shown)


pkg/tenant/db/projects_test.go
```diff
@@ -83,7 +83,7 @@ func (s *dbTestSuite) TestRetrieveProject() {
 
 	s.mock.OnGet = func(ctx context.Context, in *pb.GetRequest) (*pb.GetReply, error) {
 		if len(in.Key) == 0 || in.Namespace != db.ProjectNamespace {
-			return nil, status.Error(codes.FailedPrecondition, "bad Put request")
+			return nil, status.Error(codes.FailedPrecondition, "bad Get request")
```
bbengfort, line 86: good catch!

pkg/tenant/db/projects_test.go
```diff
@@ -162,7 +162,7 @@ func (s *dbTestSuite) TestListProjects() {
 	projects = append(projects, project)
 	require.Len(projects, 1)
 
-	_, err = db.ListMembers(ctx, project.TenantID)
+	_, err = db.ListProjects(ctx, project.TenantID)
```
bbengfort, line 165: Sorry we didn't catch that in the last review.

pkg/tenant/db/topics.go
```diff
@@ -0,0 +1,135 @@
...
+
+func (t *Topic) UnmarshalValue(data []byte) error {
+	return msgpack.Unmarshal(data, t)
+}
+
+// CreateTopic adds a new topic to the database.
+func CreateTopic(ctx context.Context, topic *Topic) (err error) {
+	if topic.ID.Compare(ulid.ULID{}) == 0 {
+		topic.ID = ulid.Make()
+	}
```
bbengfort, line 58: We should probably create a story to add data validation here, e.g. in the case that there is no `ProjectID` we should return an error; topics should also have their topic names validated; they can only contain alphanumerics and _ and cannot start with a number. 

We can create a `Validate()` method on the model that we can call before create and and update to ensure that our data structure is valid.

bbengfort (approved):
Looks good - I'm glad we're fleshing out all of the stubs for our models and utilities. Now that we've gotten this far, I'm starting to think about data validation and making sure that our database is in a correct state. For example, a topic should not be able to be saved without a ProjectID. There might be other validation that we might want to do as well -- I think we should create some follow on stories for these topics.

PR #84, opened 2023-01-05
Sc 11924/mem list

daniellemaxwell:
### Scope of changes

Adds list handlers for the members, projects, and topics resources of the Tenant API. The Tenant ID for TenantMemberList and TenantProjectList and the Project ID for ProjectTopicList are parsed from the URL.

The handlers for MemberList, ProjectList, and TopicList currently pass in an empty ULID because the Tenant and Project IDs will need to be obtained from JWT credentials that will be provided via Quarterdeck's authentication middleware. Once the middleware is created, the handlers will be updated to filter by the appropriate ID.

Fixes SC-11924

### Type of change

- [x] new feature
- [ ] bug fix
- [ ] documentation
- [x] testing
- [ ] technical debt
- [ ] other (describe)

### Acceptance criteria

### Author checklist

### Reviewer(s) checklist


pkg/tenant/members.go
```diff
@@ -10,8 +10,39 @@ import (
...
+		log.Error().Err(err).Msg("could not fetch members from the database")
+		c.JSON(http.StatusInternalServerError, api.ErrorResponse("could not fetch members from the database"))
+		return
+	}
+
+	out := &api.TenantMemberPage{TenantMembers: make([]*api.Member, 0)}
+
+	tenantMember := &api.Member{}
+
+	out.TenantMembers = append(out.TenantMembers, tenantMember)
```
rebeccabilbro, line 43: It looks like on line 33 we have the potential to obtain a slice of members (type `[]*Member`) from the database, but we're currently assigning the result to a `_`. Should we instead be capturing the result into a new var such as `members` and then assigning `out.TenantMembers = members` rather than appending an empty `tenantMember`?

pkg/tenant/members.go
```diff
@@ -67,21 +98,29 @@ func (s *Server) TenantMemberCreate(c *gin.Context) {
 	c.JSON(http.StatusCreated, member)
 }
 
+// MemberList retrieves all members assigned to a tenant
```
rebeccabilbro, line 101: I'm not sure I'm quite understanding how the `MemberList` function differs from `TenantMemberList` above -- it looks like both are designed to extract members associated with a tenant from the database. Is the only difference the route?

pkg/tenant/members.go
```diff
@@ -67,21 +98,29 @@ func (s *Server) TenantMemberCreate(c *gin.Context) {
...
 
-	// TODO: Add authentication and authorization middleware
-	// TODO: Identify top-level info
-	// TODO: Parse and validate user input
-	// TODO: Perform work on the request, e.g. database interactions,
-	// sending notifications, accessing other services, etc.
+	member := &api.Member{}
 
-	// Return response with the correct status code
+	out.Members = append(out.Members, member)
```
rebeccabilbro, line 121: I have the same question here as in the `TenantMemberList` function -- it seems like we should be using the slice of `members` the database passes back to us on line 110 instead of assigning it to `_`, and then perhaps assigning `out.Members = members` rather than appending an empty `member`

pkg/tenant/members_test.go
```diff
@@ -51,6 +80,31 @@ func (suite *tenantTestSuite) TestTenantMemberCreate() {
...
+
+	// Call the OnCursor method.
+	trtl.OnCursor = func(cr *pb.CursorRequest, t pb.Trtl_CursorServer) error {
+		return nil
+	}
+
+	req := &api.PageQuery{
+		PageSize:      2,
+		NextPageToken: "12",
+	}
```
rebeccabilbro, line 101: Just curious -- is this necessary for the test or just a rule of thumb for Ensign database calls? What happens if no pagination instructions are specified here (and same question for lines 30-33)?

pkg/tenant/members_test.go
```diff
@@ -12,6 +12,35 @@ import (
...
+
+	// Should return an error if the tenant does not exist.
+	_, err := suite.client.TenantMemberList(ctx, "invalid", req)
+	suite.requireError(err, http.StatusBadRequest, "could not parse tenant ulid", "expected error when tenant does not exist")
+
+	members, err := suite.client.TenantMemberList(ctx, "01ARZ3NDEKTSV4RRFFQ69G5FAV", req)
+	require.NoError(err, "could not list tenant members")
+	require.Len(members.TenantMembers, 1, "expected one member in the database")
```
rebeccabilbro, line 41: Do we also need to test that we get back the specific member that we're expecting (in addition to making sure we get the right # of members back)?

pkg/tenant/projects.go
```diff
@@ -10,29 +10,68 @@ import (
...
+	var tenantID ulid.ULID
+	if tenantID, err = ulid.Parse(c.Param("tenantID")); err != nil {
+		log.Error().Err(err).Msg("could not parse tenant ulid")
+		c.JSON(http.StatusBadRequest, api.ErrorResponse("could not parse tenant ulid"))
+		return
+	}
+
+	// Get projects from the database and return a 500 response
+	// if not successful.
+	if _, err := db.ListProjects(c.Request.Context(), tenantID); err != nil {
```
rebeccabilbro, line 32: It looks like `ListProjects` will hand us back a slice of projects (type `[]*Project`), which I think we want to assign to a variable rather than to a `_` so that we can capture the database response, and then assign it to `out.TenantProjects`.

pkg/tenant/projects.go
```diff
@@ -10,29 +10,68 @@ import (
...
 }
 
 func (s *Server) TenantProjectCreate(c *gin.Context) {
 	c.JSON(http.StatusNotImplemented, "not implemented yet")
 }
 
+// ProjectList retrieves all projects assigned to a tenant
+// and returns a 200 OK response.
+//
+// Route: /projects
```
rebeccabilbro, line 55: Again, I'm not totally sure I get the distinction between `TenantProjectList` and `ProjectList`, does it have to do with the specificity of the client's request? Or maybe it's a use case thing, e.g. if I have multiple tenants vs. if I only have one tenant? Perhaps we can add a bit of clarification in the docstring?

pkg/tenant/projects.go
```diff
@@ -10,29 +10,68 @@ import (
...
+	// TODO: Fetch the project's tenant ID from key.
 
-	// TODO: Add authentication and authorization middleware
-	// TODO: Identify top-level info
-	// TODO: Parse and validate user input
-	// TODO: Perform work on the request, e.g. database interactions,
-	// sending notifications, accessing other services, etc.
+	// Get projects from the database and return a 500 response
+	// if not successful.
+	if _, err := db.ListProjects(c.Request.Context(), ulid.ULID{}); err != nil {
```
rebeccabilbro, line 61: As noted above, I think we want to capture the db values passed back into a variable here which can be assigned to an attribute of `out`

pkg/tenant/topics.go
```diff
@@ -10,29 +10,68 @@ import (
...
+	var projectID ulid.ULID
+	if projectID, err = ulid.Parse(c.Param("projectID")); err != nil {
+		log.Error().Err(err).Msg("could not parse project ulid")
+		c.JSON(http.StatusBadRequest, api.ErrorResponse("could not parse project ulid"))
+		return
+	}
+
+	// Get topics from the database and return a 500 response
+	// if not successful.
+	if _, err = db.ListTopics(c.Request.Context(), projectID); err != nil {
```
rebeccabilbro, line 33: Looks like we will want to store the topics (`[]*Topic`) passed back by the database here so that we can assign the result to an `out` attribute

pkg/tenant/topics.go
```diff
@@ -10,29 +10,68 @@ import (
...
 }
 
 func (s *Server) ProjectTopicCreate(c *gin.Context) {
 	c.JSON(http.StatusNotImplemented, "not implemented yet")
 }
 
+// TopicList retrieves all topics assigned to a project and
+// returns a 200 OK response.
+//
+// Route: /topics
```
rebeccabilbro, line 56: Could we add a bit of clarification in the docstring about how this function differs from `ProjectTopicList`?

pkg/tenant/topics.go
```diff
@@ -10,29 +10,68 @@ import (
...
+	// TODO: Fetch the topic's project ID from key.
 
-	// TODO: Add authentication and authorization middleware
-	// TODO: Identify top-level info
-	// TODO: Parse and validate user input
-	// TODO: Perform work on the request, e.g. database interactions,
-	// sending notifications, accessing other services, etc.
+	// Get topics from the database and return a 500 response
+	// if not successful.
+	if _, err := db.ListTopics(c.Request.Context(), ulid.ULID{}); err != nil {
```
rebeccabilbro, line 62: Looks like we will want to store the topics (`[]*Topic`) passed back by the database here so that we can assign the result to an `out` attribute

rebeccabilbro: Also, perhaps we should create the `ulid.ULID{}` right after line 58 so that we remember that this should be a value parsed from the key rather than an empty struct in the future?

pkg/tenant/topics_test.go
```diff
@@ -12,6 +12,60 @@ import (
...
+		NextPageToken: "12",
+	}
+
+	// Should return an error if the project does not exist.
+	_, err := suite.client.ProjectTopicList(ctx, "invalid", req)
+	suite.requireError(err, http.StatusBadRequest, "could not parse project ulid", "expected error when project does not exist")
+
+	topics, err := suite.client.ProjectTopicList(ctx, "01GNA926JCTKDH3VZBTJM8MAF6", req)
+	require.NoError(err, "could not list project topics")
+	require.Len(topics.TenantTopics, 1, "expected one topic in the database")
```
rebeccabilbro, line 41: Should we add a test to ensure we get back the topic we expect?

pkg/tenant/projects_test.go
```diff
@@ -12,6 +12,60 @@ import (
...
+		NextPageToken: "12",
+	}
+
+	// Should return an error if the tenant does not exist.
+	_, err := suite.client.TenantProjectList(ctx, "invalid", req)
+	suite.requireError(err, http.StatusBadRequest, "could not parse tenant ulid", "expected error when tenant does not exist")
+
+	projects, err := suite.client.TenantProjectList(ctx, "01GKKYAWC4PA72YC53RVXAEC67", req)
+	require.NoError(err, "could not list tenant projects")
+	require.Len(projects.TenantProjects, 1, "expected one project in the database")
```
rebeccabilbro, line 41: Should we add a test to ensure we get back the project we expect?

rebeccabilbro (commented):
Thank you @daniellemaxwell for the opportunity to dig into some Ensign code; it was very exciting to see how far you guys have come with the backend in such a short amount of time! As a first pass through the code, I have made a few comments for your consideration. Since I'm coming at the code pretty fresh, there are a few things that are still a bit unfamiliar about the implementation (e.g. the functions that seem a bit similar to each other like `TenantProjectList` vs `ProjectList`), so forgive me if some of my questions seem a bit naive!

pkg/tenant/members.go
daniellemaxwell: Yes, currently the difference between the two is that the route is different. If we were to assign a member to an organization, `MemberList` could be updated and then provide a distinction between the two.

pkg/tenant/members_test.go
daniellemaxwell: A `PageQuery` must be passed in for all of the list functions, but this will work if pagination instructions aren't included.

pkg/tenant/members.go
daniellemaxwell: Line 33 obtains a slice of `[]*db.Member` and this cannot be assigned to `out.TenantMembers`. I've attempted a few different things, following your suggestion, but can't quite get anything to work along those lines. I'll give this another try on Monday and I haven't made any progress.

pkg/tenant/members_test.go
daniellemaxwell: I'll look into this more on how to add that test.

pkg/tenant/projects.go
daniellemaxwell: We want `ProjectList` to provide a list of projects for an organization and `TenantProjectList` will list projects on the tenant. I'll update the docstring.

pkg/tenant/members.go
daniellemaxwell: There's been a bit of confusion on my side about this as well, so I want to follow up after looking back at a discussion about the difference. `MemberList` will list members for an organization and `TenantMemberList` will list members for a tenant. The confusing part for me came about because the Member model has a tenant ID instead of an object ID. However, the tenant ID will also be stored on the organization so that will aid in making that distinction. I'll update the docstring.

pkg/tenant/api/v1/api.go
```diff
@@ -132,7 +132,7 @@ type ProjectPage struct {
 
 type ProjectTopicPage struct {
 	ProjectID     string `json:"project_id"`
-	TenantTopics  []*Topic
+	Topics        []*Topic
```
daniellemaxwell, line 135: Corrected the name since there will only be one type of topic.

pkg/tenant/db/members.go
```diff
@@ -100,10 +99,10 @@ func ListMembers(ctx context.Context, tenantID ulid.ULID) (members []*Member, er
 	members = make([]*Member, 0, len(values))
 	for _, data := range values {
 		member := &Member{}
-		if err = member.UnmarshalValue(data); err != nil {
-			fmt.Println(err)
-			return nil, err
-		}
+
+		// Marshal and unmarshal the data with msgPack.
+		member.MarshalPackData()
```
daniellemaxwell, line 104: The list handlers were returning empty arrays because the data ( a `[]byte`) needs to be marshaled and unmarshaled by msgPack. Previously, it was only unmarshaled. A method was created to make this a little cleaner.

pkg/tenant/db/members_test.go
```diff
@@ -148,14 +148,11 @@ func (s *dbTestSuite) TestListMembers() {
...
 
-	members := make([]*db.Member, 0, len(values))
-	members = append(members, member)
-	require.Len(members, 1)
-
-	_, err = db.ListMembers(ctx, member.TenantID)
-	require.Error(err, "could not list members")
+	members, err := db.ListMembers(ctx, member.TenantID)
+	require.NoError(err, "could not list members")
+	require.Len(members, 7, "expected 7 members")
```
pdeziel, line 155: I like that we're checking the length of the returned list, we should also check that one of the values in the list is correct (or ideally all of them) to make sure the unmarshaling is working correctly.

pkg/tenant/members.go
```diff
@@ -10,8 +10,50 @@ import (
...
+
+	// Build the response.
+	out := &api.TenantMemberPage{
+		TenantID:      tenantID.String(),
+		TenantMembers: make([]*api.Member, 0),
+	}
+
+	// Loop over member. For each db.Member inside the array, create a tenantMember
+	// which will be an api.Member{} and assign the ID and Name fetched from db.Member
+	// to that struct and then append to the out.TenantMembers array.
```
pdeziel, line 48: Nice! Glad you were able to figure this bit out.

pkg/tenant/members.go
```diff
@@ -83,21 +125,30 @@ func (s *Server) TenantMemberCreate(c *gin.Context) {
...
 
-	// TODO: Add authentication and authorization middleware
-	// TODO: Identify top-level info
-	// TODO: Parse and validate user input
-	// TODO: Perform work on the request, e.g. database interactions,
-	// sending notifications, accessing other services, etc.
+	member := &api.Member{}
 
-	// Return response with the correct status code
+	out.Members = append(out.Members, member)
```
pdeziel, line 149: So do we need to do the same thing here that we did in `TenantMemberList`? Loop through each member model and construct the `api.Member`?

pkg/tenant/members_test.go
```diff
@@ -1,17 +1,68 @@
...
+			stream.Send(&pb.KVPair{
+				Key:       []byte(fmt.Sprintf("key %d", i)),
+				Value:     []byte(fmt.Sprintf("value %d", i)),
+				Namespace: in.Namespace,
+			})
+		}
+		return nil
+	}
+
+	// Should return an error if the tenant does not exist.
```
pdeziel, line 57:
```suggestion
	// Should return an error if the tenant ID is missing.
```

pkg/tenant/members_test.go
```diff
@@ -1,17 +1,68 @@
...
+	_, err := suite.client.TenantMemberList(ctx, "invalid", &api.PageQuery{})
+	suite.requireError(err, http.StatusBadRequest, "could not parse tenant ulid", "expected error when tenant does not exist")
```
pdeziel, line 59:
```suggestion
	suite.requireError(err, http.StatusBadRequest, "could not parse tenant ulid", "expected error when tenant ID is missing")
```

pkg/tenant/members_test.go
```diff
@@ -1,17 +1,68 @@
...
+
+	members, err := suite.client.TenantMemberList(ctx, member.TenantID.String(), &api.PageQuery{})
+	require.NoError(err, "could not list tenant members")
+	require.Len(members.TenantMembers, 7, "expected 7 members")
```
pdeziel, line 63: Similar to the other test, we should make sure that the unmarshaling is happening correctly by checking each value in the list here.

pkg/tenant/projects.go
```diff
@@ -76,21 +118,30 @@ func (s *Server) TenantProjectCreate(c *gin.Context) {
...
 
-	// TODO: Add authentication and authorization middleware
-	// TODO: Identify top-level info
-	// TODO: Parse and validate user input
-	// TODO: Perform work on the request, e.g. database interactions,
-	// sending notifications, accessing other services, etc.
+	project := &api.Project{}
 
-	// Return response with the correct status code
+	out.Projects = append(out.Projects, project)
```
pdeziel, line 142: Similar to `MemberList` we need to populate the list of returned `api.Project` here.

pdeziel (commented):
Looks like we're getting closer! It sounds like we just need to shore up the unmarshaling and some of the other endpoint methods. I also had a suggestion for improving the tests a bit once the unmarshaling bit is figured out.

pkg/tenant/members.go
daniellemaxwell: A follow on story was created to filter the list by the `Tenant ID` once the authentication middleware was available, so I decided to make the change you've mentioned here when I work on that. Now that the middleware has been created, I'll work on that soon.

pkg/tenant/projects.go
daniellemaxwell: Same as the comment above about `MemberList`.

pdeziel (approved):
Looks good to me assuming you have creating the follow up stories. Thanks for sticking with it!

PR #99, opened 2023-01-20
APIKey Model Update

bbengfort:
### Scope of changes

Updates APIKey Database Model in preparation for API Keys endpoints.

Fixes SC-12824

### Type of change

- [x] new feature
- [ ] bug fix
- [ ] documentation
- [x] testing
- [ ] technical debt
- [ ] other (describe)

### Acceptance criteria

Please especially review the changes that have been made to the API models for integration with Tenant.


pdeziel (approved):
Looks good to me!

pkg/quarterdeck/api/v1/api.go
```diff
@@ -114,15 +116,19 @@ type APIAuthentication struct {
...
-	Permissions  []string `json:"permissions,omitempty"`
-	Created      string   `json:"created,omitempty"`
-	Modified     string   `json:"modified,omitempty"`
+	ID           ulid.ULID `json:"id,omitempty"`            // not allowed on create
+	ClientID     string    `json:"client_id"`               // not allowed on create, cannot be updated
+	ClientSecret string    `json:"client_secret,omitempty"` // not allowed on created, cannot be updated
+	Name         string    `json:"name"`                    // required on create, update
+	OrgID        ulid.ULID `json:"org_id"`                  // required on create, cannot be updated
+	ProjectID    ulid.ULID `json:"project_id"`              // required on create, cannot be updated
+	CreatedBy    ulid.ULID `json:"owner,omitempty"`         // required on create, cannot be updated
```
pdeziel, line 125: Is the `CreatedBy` datatype correct? It's listed as `sql.NullByte` in the database model instead of `ulid.ULID`

pkg/quarterdeck/db/models/apikeys.go
```diff
@@ -66,6 +74,110 @@ func GetAPIKey(ctx context.Context, clientID string) (key *APIKey, err error) {
...
+// timestamps even if the user has already set them on the model. If the APIKey does not
+// have a client ID and secret, they're generated before the model is created.
+func (k *APIKey) Create(ctx context.Context) (err error) {
+	k.ID = ulids.New()
+
+	now := time.Now()
+	k.SetCreated(now)
+	k.SetModified(now)
+
+	if k.KeyID == "" && k.Secret == "" {
```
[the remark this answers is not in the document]
bbengfort: Yes, I'm hoping that situation doesn't happen!

pkg/quarterdeck/api/v1/api.go
bbengfort: `sql.NullByte` is the only way we can have a nullable ULID -- the database storage for a ULID is `BLOB` (bytes) - the `ULID` is a `[16]byte` so it automatically gets converted to `[]byte` when going into and out of the database and a type conversion happens when it is retrieved. If we used `ulid.ULID` for the nullable field, then we would always insert `0x00 x16` into the database. 

I guess the question is if an API Key can be created without an owner account -- if not, we could always make it NOT NULL.

PR #102, opened 2023-01-20
Filter list handlers by id

daniellemaxwell:
### Scope of changes

This will allow list handlers for Tenant API resources to filter by their organization ID.

Fixes SC-12216

### Type of change

- [ ] new feature
- [ ] bug fix
- [ ] documentation
- [ ] testing
- [ ] technical debt
- [ ] other (describe)

### Acceptance criteria

### Author checklist

### Reviewer(s) checklist

### DIFF
--- a/pkg/tenant/members.go
+++ b/pkg/tenant/members.go
@@ -5,6 +5,8 @@ import (
 
 	"github.com/gin-gonic/gin"
 	"github.com/oklog/ulid/v2"
+	middleware "github.com/rotationalio/ensign/pkg/quarterdeck/middleware"
+	"github.com/rotationalio/ensign/pkg/quarterdeck/tokens"
 	"github.com/rotationalio/ensign/pkg/tenant/api/v1"
 	"github.com/rotationalio/ensign/pkg/tenant/db"
 	"github.com/rs/zerolog/log"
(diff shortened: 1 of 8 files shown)


pkg/tenant/members.go
```diff
@@ -131,12 +133,29 @@ func (s *Server) TenantMemberCreate(c *gin.Context) {
...
-	// if not succesful.
-	if _, err := db.ListMembers(c.Request.Context(), tenantID); err != nil {
+	// Fetch member from the context.
+	if member, err = middleware.GetClaims(c); err != nil {
+		log.Error().Err(err).Msg("could not fetch member from context")
+		c.JSON(http.StatusInternalServerError, api.ErrorResponse("could not fetch member from context"))
+		return
+	}
+
+	// Get the member's orgnaization ID and return a 400 response if it is not a ULID.
```
pdeziel, line 148: I am wondering if we should return a 401 response here if the `member.OrgID` is empty. Since we are retrieving the `OrgID` from the claims, if the field is empty then it implies that something went wrong with authentication instead of something in the request itself (e.g. there's a middleware bug or the frontend sent a bad token with the request).

Similarly, perhaps if the OrgID is not empty but is not a ULID should we return a 500?

pkg/tenant/members_test.go
```diff
@@ -198,19 +198,67 @@ func (suite *tenantTestSuite) TestMemberList() {
...
+			})
+		}
 		return nil
 	}
 
 	// Set the initial claims fixture
 	claims := &tokens.Claims{
 		Name:        "Leopold Wentzel",
 		Email:       "leopold.wentzel@gmail.com",
+		OrgID:       "01GMTWFK4XZY597Y128KXQ4WHP",
```
pdeziel, line 261: It might also be interesting to have a test case where the `OrgID` is missing from the claims or is not a ULID, to make sure we're returning the right status code in that case. The `suite.SetClientCredentials()` should help with that.

pkg/tenant/members_test.go
```diff
@@ -227,9 +275,24 @@ func (suite *tenantTestSuite) TestMemberList() {
...
+
+	// Verify second member data has been populated.
+	require.Equal(members[1].ID.String(), rep.Members[1].ID, "member id should match")
+	require.Equal(members[1].Name, rep.Members[1].Name, "member name should match")
+	require.Equal(members[1].Role, rep.Members[1].Role, "member role should match")
+
+	// Verify third member data has been populated.
+	require.Equal(members[2].ID.String(), rep.Members[2].ID, "member id should match")
+	require.Equal(members[2].Name, rep.Members[2].Name, "member name should match")
+	require.Equal(members[2].Role, rep.Members[2].Role, "member role should match")
```
pdeziel, line 295: This is a great test. One thing you can also do to reduce the repeated code here is to just make it a loop, e.g.

```golang
for i, _ := range members {
    require.Equal(members[i].ID.String(), rep.Members[i].ID, "member id should match")
    ...
}
```

This improves the maintainability a little bit so that if for example we add more fields to the members response in the future we only have to add one line of code instead of three.

Just something to think about but not a necessary change for merging this PR.

pdeziel (approved):
This looks great, I'm glad that we're keeping up with the test coverage! I had some minor comments about status codes but nothing that should block merging this in.

pkg/tenant/members.go
daniellemaxwell: I'd debated over whether a 500 would be more appropriate if the OrgID isn't a ULID, so I agree with you here and will update accordingly.

pkg/tenant/members_test.go
daniellemaxwell: I kept getting a `ulid bad data size` error before realizing the OrgID needed to be added here, so I'm sure adding the test mentioned will be very helpful.

PR #111, opened 2023-01-23
Add authentication middleware to create handlers

daniellemaxwell:
### Scope of changes

Adds authentication middleware to fetch the organization ID for `TenantCreate`, `MemberCreate`, and `ProjectCreate`. Updated the claims test fixture with an OrgID and tested error for when the OrgID is missing or isn't a valid ULID.

Note: A topic should not be created for an organization, only for a project.

Fixes SC-12946

### Type of change

- [x] new feature
- [ ] bug fix
- [ ] documentation
- [x] testing
- [ ] technical debt
- [ ] other (describe)

### Acceptance criteria

### Author checklist

### Reviewer(s) checklist


pkg/tenant/members.go
```diff
@@ -184,11 +184,23 @@ func (s *Server) MemberList(c *gin.Context) {
...
-	// TODO: Add authentication middleware to fetch the organization ID.
+	// Fetch member from the context.
+	if m, err = middleware.GetClaims(c); err != nil {
+		log.Error().Err(err).Msg("could not fetch member from context")
+		c.JSON(http.StatusUnauthorized, api.ErrorResponse(err))
+		return
+	}
+
+	// Get the member's organization ID and return a 500 response if it is not a ULID.
+	if _, err = ulid.Parse(m.OrgID); err != nil {
```
daniellemaxwell, line 199: The Member model takes in a Tenant ID, not an OrgID and that is why the value isn't captured here and used elsewhere.

pkg/tenant/projects.go
```diff
@@ -175,11 +175,23 @@ func (s *Server) ProjectList(c *gin.Context) {
...
 
-	// TODO: Add authentication middleware to fetch the organization ID.
+	// Fetch project from the context.
+	if p, err = middleware.GetClaims(c); err != nil {
+		log.Error().Err(err).Msg("could not fetch project from context")
+		c.JSON(http.StatusUnauthorized, api.ErrorResponse(err))
+		return
+	}
+
+	// Get the project's organization ID and return a 500 response if it is not a ULID.
```
daniellemaxwell, line 189: The Project model takes in a Tenant ID, not an OrgID and that is why the value isn't captured here and used elsewhere.

pkg/tenant/members.go
pdeziel: I see, since this isn't going to work for right now (the members and projects models need to be fleshed out as you pointed out) can we add a check at the top of this method and in `ProjectCreate` so that this returns not implemented?

```golang
	if pkg.VersionReleaseLevel == "alpha" {
		c.JSON(http.StatusNotImplemented, api.ErrorResponse("not implemented"))
		return
	}
```

Then once we're past the alpha release we can go back and fix these. Also we should probably add a `suite.T().Skip()` to the corresponding tests so they are skipped for now.

pkg/tenant/tenants.go
```diff
@@ -67,54 +67,69 @@ func (s *Server) TenantList(c *gin.Context) {
...
-		err error
-		t   *api.Tenant
+		err    error
+		t      *tokens.Claims
+		tenant *api.Tenant
 	)
 
-	// TODO: Add authentication and authorization middleware
+	// Fetch tenant from the context.
+	if t, err = middleware.GetClaims(c); err != nil {
```
pdeziel, line 76: I tend to name these variables `claims` or similar, since it refers to user claims and doesn't really have anything to do with a tenant (expect that it contains an orgid). That will probably help reduce confusion with the two other "tenant" variables in this method.

pkg/tenant/tenants.go
```diff
@@ -67,54 +67,69 @@ func (s *Server) TenantList(c *gin.Context) {
...
+		log.Error().Err(err).Msg("could not fetch tenant from context")
+		c.JSON(http.StatusUnauthorized, api.ErrorResponse(err))
+		return
+	}
+
+	// Get the tenant's organization ID and return a 500 response if it is not a ULID.
+	var orgID ulid.ULID
+	if orgID, err = ulid.Parse(t.OrgID); err != nil {
+		log.Error().Err(err).Msg("could not parse org id")
```
pdeziel, line 85: One helpful thing for future debugging might be to add a string field to the log call with the orgid we tried to parse. That way it shows up in the logs if it was a badly formatted ULID.

e.g. `log.Error().Err(err).Str("orgid", t.OrgID).Msg("could not parse org id")`

pdeziel (changes requested):
Looks good, I had a comment about adding checks to `MemberCreate` and `ProjectCreate` so that they return not implemented until we're in a better state to deal with them. For the alpha release, they shouldn't be required because we have the `TenantMemberCreate` and `TenantProjectCreate` endpoints which users will actually be using.

daniellemaxwell:
@pdeziel The requested changes have been made. I'll create follow on stories to add OrgID to the Member and Project models along with stories to update the handlers.

pdeziel (approved):
Thanks for making the changes!

PR #115, opened 2023-01-24
Implemented refresh endpoint in quarterdeck

pdamodaran:
### Scope of changes

Implemented Refresh Endpoint in `auth.go` in quarterdeck.  The change includes changing function signatures in `api.go` and `client.go`.  I have also added tests for the new functionality.

Fixes SC-12843

### Type of change

- [x] new feature
- [ ] bug fix
- [ ] documentation
- [ ] testing
- [ ] technical debt
- [ ] other (describe)

### Acceptance criteria

### Author checklist

### Reviewer(s) checklist


pkg/quarterdeck/auth.go
```diff
@@ -254,7 +254,60 @@ func (s *Server) Authenticate(c *gin.Context) {
...
+	}
+
+	// get the user from the database using the ID
+	user, err := models.GetUser(c, claims.Subject)
+	if err != nil {
+		log.Error().Err(err).Msg("could not retrieve user from claims")
+		c.JSON(http.StatusUnauthorized, api.ErrorResponse("could not retrieve user from claims"))
+		return
+	}
+
```
bbengfort, line 294: I'm glad we're getting the user from the database, that's a good check to make sure that the user still exists or hasn't been deleted (we do not want to keep authenticating a deleted user). I'm thinking that we should also fetch the claims from the database again rather than use the claims from the refresh token; this will ensure that if the user's permissions (or name or email) have changed between login and refresh that those new claims are what gets sent rather than the old claims. 

Are there any other checks or security related things you can think of that we should do in this PR?

pkg/quarterdeck/auth.go
```diff
@@ -254,7 +254,60 @@ func (s *Server) Authenticate(c *gin.Context) {
...
+	// create a new access token/refresh token pair
+	out = &api.LoginReply{}
+	if out.AccessToken, out.RefreshToken, err = s.tokens.CreateTokenPair(claims); err != nil {
+		log.Error().Err(err).Msg("could not create jwt tokens on refresh")
+		c.JSON(http.StatusInternalServerError, api.ErrorResponse("could not create credentials"))
+		return
+	}
+
+	// Update the users last login in a Go routine so it doesn't block
+	// TODO: create a channel and workers to update last login to limit the num of go routines
```
bbengfort, line 304: I'll create a story for this TODO.

bbengfort: See SC-13108

pkg/quarterdeck/auth.go
```diff
@@ -254,7 +254,60 @@ func (s *Server) Authenticate(c *gin.Context) {
 	c.JSON(http.StatusOK, out)
 }
 
-// TODO: implement the refresh endpoint
+// Refresh is oriented to machine users that have a refresh token and need to obtain
+// a new access token and refresh token to continue using ensign systems
```
bbengfort, line 258:
```suggestion
// Refresh re-authenticates users and api keys using a refresh token rather than requiring a username
// and password or API key credentials a second time and returns a new access and refresh token pair
// with the current credentials of the user. This endpoint is intended to facilitate long-running
// connections to ensign systems that last longer than the duration of an access token; e.g. long
// sessions on the Beacon UI or (especially) long running publishers and subscribers (machine users)
// that need to stay authenticated semi-permanently. 
```

bbengfort (approved):
This looks great! I hope that doing a complete end-to-end endpoint with the various stubs in place made it a bit clearer how we organize our REST APIs and has given you a clearer picture of how to add new endpoints in the future. I think the only thing that was missing from this particular PR was adding a route in quarterdeck.go since it was already added. 

I had a couple of comments and minor suggestions. The major suggestion was reloading the claims from the database rather than simply reusing the claims from the refresh token. If you'd like to do that in this PR, that's fine -- otherwise go ahead and create a story for making that change. 

Also, could we create a story to fix the issue that you identified yesterday? Namely that both access and refresh tokens can be used to refresh? We'd like to make it so that only refresh tokens can be used. I think the way to deal with that is to have a different audience on the refresh token.

pdamodaran:
@bbengfort - thanks for your review!  I will implement the change to load the claims from the database and ping you for another review.

bbengfort (approved):
Looks great, thank you for making those changes! Since you marked it as the reviewer can merge, I'll go ahead and do this so this story will be moved to done in shortcut!

PR #116, opened 2023-01-24
Detail and delete api key tenant handlers

pdeziel:
### Scope of changes

This implements the detail and delete api key tenant handlers. These should be pretty straightforward given the pattern setup in SC-10407.

There was also a slight change to `ProjectAPIKeyList` which should also be reviewed.

Fixes SC-10411

### Type of change

- [x] new feature
- [ ] bug fix
- [ ] documentation
- [ ] testing
- [ ] technical debt
- [ ] other (describe)

### Acceptance criteria

The endpoint handlers make sense given the TODOs in the code.

### Author checklist

### Reviewer(s) checklist

### DIFF
--- a/pkg/tenant/apikeys.go
+++ b/pkg/tenant/apikeys.go
@@ -239,5 +239,5 @@ func (s *Server) APIKeyDelete(c *gin.Context) {
 		return
 	}
 
-	c.JSON(http.StatusNoContent, nil)
+	c.Status(http.StatusNoContent)
 }
--- a/pkg/tenant/apikeys.go
+++ b/pkg/tenant/apikeys.go
@@ -49,6 +49,7 @@ func (s *Server) ProjectAPIKeyList(c *gin.Context) {
 	}
 
 	// Request a page of API keys from Quarterdeck
+	// TODO: Handle error status codes returned by Quarterdeck
 	var reply *qd.APIKeyList
 	if reply, err = s.quarterdeck.APIKeyList(ctx, req); err != nil {
 		log.Error().Err(err).Msg("could not list API keys")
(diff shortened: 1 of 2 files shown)


pkg/tenant/apikeys.go
```diff
@@ -65,13 +66,9 @@ func (s *Server) ProjectAPIKeyList(c *gin.Context) {
...
-			ID:          key.ID.String(),
-			ClientID:    key.ClientID,
-			Name:        key.Name,
-			Owner:       key.CreatedBy.String(),
-			Permissions: key.Permissions,
-			Created:     key.Created.Format(time.RFC3339Nano),
-			Modified:    key.Modified.Format(time.RFC3339Nano),
+			ID:       key.ID.String(),
+			ClientID: key.ClientID,
+			Name:     key.Name,
```
pdeziel, line 71: List is only supposed to return some of the fields (see the quarterdeck implementation) so I've updated the handler to reflect that.

pkg/tenant/apikeys.go
```diff
@@ -168,14 +166,78 @@ func (s *Server) APIKeyCreate(c *gin.Context) {
...
+		c.JSON(http.StatusUnauthorized, api.ErrorResponse("could not fetch credentials for authenticated user"))
+		return
+	}
+
+	// Parse the API key ID from the URL
+	apiKeyID := c.Param("apiKeyID")
+
+	// Delete the API key using Quarterdeck
+	// TODO: Handle error status codes returned by Quarterdeck
+	if err = s.quarterdeck.APIKeyDelete(ctx, apiKeyID); err != nil {
```
pdeziel, line 236: Note that quarterdeck will actually return 404 if the API key is not in the user's organization (to prevent users deleting keys from other organizations). However, we should probably be handling the different error cases that get returned from Quarterdeck so I've added some TODOs in this file.

pkg/tenant/apikeys.go
```diff
@@ -168,14 +166,78 @@ func (s *Server) APIKeyCreate(c *gin.Context) {
...
+		log.Error().Err(err).Str("apiKeyID", apiKeyID).Msg("could not delete API key")
+		c.JSON(http.StatusInternalServerError, api.ErrorResponse("could not delete API key"))
+		return
+	}
+
+	c.JSON(http.StatusNoContent, nil)
```
daniellemaxwell, line 242: Could this be replaced with `c.Status(http.StatusNoContent)` or does nil need to be returned here?

daniellemaxwell (approved):
Thank you for providing comments in the PR! I have one small question, but this is good to go.

pkg/tenant/apikeys.go
pdeziel: Yes, thanks for catching this!

PR #152, opened 2023-02-02
Add projectID to quarterdeck register

pdeziel:
### Scope of changes

This changes quarterdeck's register endpoint to accept a project ID. This is currently a bit of a workaround to allow Tenant to create default resources for the user on its Register endpoint. Quarterdeck now requires the projectID as part of the register URL and creates an association between the user's new organization and the project ID, which allows the user to access it when they login since the orgID will be present in their claims.

Fixes SC-13495

### Type of change

- [x] new feature
- [ ] bug fix
- [ ] documentation
- [ ] testing
- [ ] technical debt
- [ ] other (describe)

### Acceptance criteria

### Author checklist

### Reviewer(s) checklist


bbengfort (commented):
Adding the project ID to the URL is an interesting idea for the registration endpoint, and I think it could work and fit the RESTful paradigm a bit better if we had a `/v1/projects/:projectID/register` endpoint in addition to the `/v1/register` endpoint. However, in order to future proof changes that we make to the front-end (for example, not creating a default project for the client); I propose that we make the `projectID` part of the `RegisterRequest` and make it optional rather than required.

pkg/quarterdeck/auth.go
```diff
@@ -81,6 +95,19 @@ func (s *Server) Register(c *gin.Context) {
...
+	// cannot issue requests with this project ID.
+	op := &models.OrganizationProject{
+		OrgID:     org.ID,
+		ProjectID: projectID,
+	}
+	if err = op.Save(ctx); err != nil {
+		log.Error().Err(err).Msg("could not link organization to project")
+		c.JSON(http.StatusInternalServerError, api.ErrorResponse("could not process registration"))
+		return
+	}
```
bbengfort, line 109: Note that this executes in a separate transaction from `user.Create`. I'd like to make the `ProjectID` optional - so having a separate transaction is perfectly fine; but it can create the possibility that a user/organization is created but an `OrganizationProject` is not created.

pkg/quarterdeck/mock/quarterdeck.go
```diff
@@ -64,7 +64,7 @@ func (s *Server) routeRequest(w http.ResponseWriter, r *http.Request) {
 	case path == StatusEP:
 		s.handlers[path](w, r)
 	case path == RegisterEP:
-		s.handlers[path](w, r)
+		s.handlers[RegisterEP](w, r)
```
bbengfort, line 67:
```suggestion
		s.handlers[path](w, r)
```

Is this stray from the changes to the domain?

bbengfort (approved):
Thank you for making that change; hopefully that makes it a bit easier to future-proof Quarterdeck!

PR #161, opened 2023-02-03
Invoke CreateProject from Tenant

pdeziel:
### Scope of changes

This completes the project security checks on the Tenant side by telling Quarterdeck to create the organization-project link when the user creates a project. This ensures that users will be able to edit api keys only in their organization since Quarterdeck can check to make sure that the organization-project mapping already exists.

Fixes SC-13122

### Type of change

- [ ] new feature
- [ ] bug fix
- [ ] documentation
- [ ] testing
- [ ] technical debt
- [x] security

### Acceptance criteria

### Author checklist

### Reviewer(s) checklist


pkg/tenant/projects.go
```diff
@@ -122,16 +132,17 @@ func (s *Server) TenantProjectCreate(c *gin.Context) {
...
 
-	// Add project to the database and return a 500 response if it cannot be added.
-	if err = db.CreateTenantProject(c.Request.Context(), tproject); err != nil {
-		log.Error().Err(err).Msg("could not create tenant project in the database")
-		c.JSON(http.StatusInternalServerError, api.ErrorResponse("could not add tenant project"))
+	// Create the project in the database and register it with Quarterdeck.
+	// TODO: Distinguish between trtl errors and quarterdeck errors.
+	if err = s.createProject(ctx, tproject); err != nil {
+		log.Error().Err(err).Msg("could not create project")
+		c.JSON(qd.ErrorStatus(err), api.ErrorResponse("could not create project"))
```
bbengfort, line 139: If this is a `trtl` error what will `qd.ErrorStatus(err)` do - does it return a 500?

bbengfort (approved):
Looks good to me. 

If the Quarterdeck project create fails do we want to delete the project in trtl database, or do we want to retry Quarterdeck project create later?

pkg/tenant/projects.go
pdeziel: I will return a 500. I was thinking we might want to have a struct similar to `StatusError` either in the tenant db package or in trtl so that we could switch on on the type here and return a more useful status code.

pdeziel:
> Looks good to me.
> 
> If the Quarterdeck project create fails do we want to delete the project in trtl database, or do we want to retry Quarterdeck project create later?

Good question - my current thinking is that we should just try to delete the project.

bbengfort:
> > Looks good to me.
> > If the Quarterdeck project create fails do we want to delete the project in trtl database, or do we want to retry Quarterdeck project create later?
> 
> Good question - my current thinking is that we should just try to delete the project.

Ok - deleting makes sense; do we have a story for that?

pkg/tenant/projects.go
bbengfort: Makes sense to me - would you mind creating a follow on story for that?

PR #173, opened 2023-02-06
fix beacon production env

pdeziel:
### Scope of changes

This hopefully fixes the production environment variables for the automated builds of the beacon UI.

Fixes SC-XXXX

### Type of change

- [ ] new feature
- [ ] bug fix
- [ ] documentation
- [ ] testing
- [ ] technical debt
- [x] docker

### Acceptance criteria

### Author checklist

### Reviewer(s) checklist

### DIFF
--- a/containers/beacon/Dockerfile
+++ b/containers/beacon/Dockerfile
@@ -13,8 +13,8 @@ COPY web/beacon-app/yarn.lock ./
 RUN yarn install --frozen-lockfile --check-files
 
 # Set environment variables
-ARG REACT_APP_TENANT_BASE_URL
-ARG REACT_APP_QUARTERDECK_BASE_URL
+ARG REACT_APP_TENANT_BASE_URL="https://api.rotational.app/v1/"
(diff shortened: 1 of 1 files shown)


bbengfort (approved):
Yes, this works for me, everything looks correct; in terms of Docker syntax, and I'm pretty sure the `/v1/` path is also required -- that would match what the front-end is doing in other places.

For the other args such as `REACT_APP_ANALYTICS_ID`; those need to be set in two places:

1. They should be added as build args to the workflow using secrets: https://github.com/rotationalio/ensign/blob/main/.github/workflows/containers.yaml#L73
2. They should be added to GitHub secrets

See this as a reference:

https://github.com/trisacrypto/directory/blob/main/.github/workflows/containers.yaml#L418-L424

Let me know if you need me to set any secrets in the repo (step 2). I presume I will. It's also important to know which env vars the front end is using. They usually have a `.env.template` in the repo, but I don't see one; it might be worth following up with @masskoder about that.

PR #192, opened 2023-02-08
Contacts data structures

pdeziel:
### Scope of changes

This story sets up the contacts data structure in the `sendgrid` package so that it can be used both when sending emails and adding contacts to marketing lists.

Fixes SC-13281

### Type of change

- [ ] new feature
- [ ] bug fix
- [ ] documentation
- [ ] testing
- [ ] technical debt
- [ ] other (describe)

### Acceptance criteria

This is mostly a design story - does it make sense to use the `Contact` struct for both generating emails and the marketing lists?

### Author checklist

### Reviewer(s) checklist

- [ ] Does the `Contact` struct make sense for both building emails and managing marketing lists and is it easy enough to construct/use from the caller perspective?


pkg/utils/emails/builders.go
```diff
@@ -54,17 +55,15 @@ const (
...
-// EmailData includes data which is required for all emails
-type EmailData struct {
-	SenderName     string
-	SenderEmail    string
-	RecipientName  string
-	RecipientEmail string
+// Contacts includes the sender and recipient of an email
+type Contacts struct {
+	Sender    sendgrid.Contact
+	Recipient sendgrid.Contact
```
bbengfort, line 61: I'm trying to think about how this data structure might evolve in the future; will we always just have one sender and one recipient or will we want multiple recipients? If so, will multiple recipients be on the same email or will a new email be created for each of them? 

Because of these questions - I think I actually prefer the `EmailData` struct name better; perhaps the `Subject` could also be in `EmailData`? That might be a more evolvable way of dealing with the information needed to send an email in the future.

bbengfort (approved):
Looks good - I like that we're unifying the sendgrid and email contacts; will make management easier and may allow us to even look up emails in sendgrid in the future.

PR #219, opened 2023-02-15
Project One Time Access Claims

bbengfort:
### Scope of changes

Implements an endpoint in Quarterdeck to issue a user one-time access to an Ensign project.

Fixes SC-13958

### Type of change

- [x] new feature
- [ ] bug fix
- [ ] documentation
- [ ] testing
- [ ] technical debt
- [ ] other (describe)

### Acceptance criteria

See checklist below.

### Author checklist

### Reviewer(s) checklist

- [ ] Is the expiration of the claim long enough/short enough?
- [ ] Does it make sense to only copy over the topic related permissions?
- [ ] Are there any claims missing in the response?
- [ ] Is it ok to reuse the login reply struct?


pkg/quarterdeck/orgs.go
```diff
@@ -113,3 +119,80 @@ func (s *Server) ProjectCreate(c *gin.Context) {
...
+		},
+		OrgID:       claims.OrgID,
+		ProjectID:   project.ProjectID.String(),
+		Permissions: make([]string, 0, 4),
+	}
+
+	// Add only the user permissions related to topics to these claims -- whatever
+	// access to topics the user has, so to will the one time access claims.
+	for _, permission := range claims.Permissions {
+		if strings.HasPrefix(permission, "topics") {
```
pdeziel, line 184: Should we define the permission prefixes in code to protect ourselves from future permissions changes?

pkg/quarterdeck/orgs_test.go
```diff
@@ -104,3 +106,69 @@ func (s *quarterdeckTestSuite) TestProjectCreate() {
...
+
+	ctx, cancel := context.WithTimeout(context.Background(), 30*time.Second)
+	defer cancel()
+
+	// Requesting one-time access to a project requires authentication
+	req := &api.Project{ProjectID: ulid.MustParse("01GQ7P8DNR9MR64RJR9D64FFNT")}
+	_, err := s.client.ProjectAccess(ctx, req)
+	s.CheckError(err, http.StatusUnauthorized, "this endpoint requires authentication")
+
+	// Requesting one-time access to a project an Project requires the topics:read permission
```
pdeziel, line 122:
```suggestion
	// Requesting one-time access to a project requires the topics:read permission
```

pdeziel (approved):
Makes sense to me, I think reusing the `LoginReply` is fine and the claim expiration also makes sense. I was trying to think if this could be abused by a user to achieve more permissions than they have but I see we're only copying over the ones that already exist in the claims, so looks good! I had a few small comments but it shouldn't block the merge.

pkg/quarterdeck/orgs.go
bbengfort: Good idea

PR #221, opened 2023-02-15
Tenant topics configuration

pdeziel:
### Scope of changes

This adds a "topics" configuration to Tenant so that it can eventually manage pub/sub event streams with Ensign. We intend to use this to create a subscriber within Tenant to subscribe to the list of topics that Ensign creates/modifies/deletes.

Fixes SC-13883

### Type of change

- [x] new feature
- [ ] bug fix
- [ ] documentation
- [ ] testing
- [ ] technical debt
- [ ] other (describe)

### Acceptance criteria

Please review the configuration to make sure it's complete and it's understandable how it's intended to be used.

### Author checklist

### Reviewer(s) checklist


pkg/tenant/config/config.go
```diff
@@ -34,6 +35,7 @@ type Config struct {
 	Auth         AuthConfig          `split_words:"true"`
 	Database     DatabaseConfig      `split_words:"true"`
 	Ensign       EnsignConfig        `split_words:"true"`
+	Topics       SDKConfig           `split_words:"true"`
```
pdeziel, line 38: If Tenant needs more event streams in the future, my thinking was that we would just add more of these.

pkg/tenant/config/config.go
```diff
@@ -71,6 +73,12 @@ type EnsignConfig struct {
 	PoolPath string `split_words:"true"`
 }
 
+// Configures an SDK connection to Ensign for pub/sub.
+type SDKConfig struct {
+	Ensign  ensign.Options
+	TopicID string `split_words:"true"`
```
pdeziel, line 79: Is this too tied to the SDK?

bbengfort: Does the `ensign.Options` have `envconfig` variables?

bbengfort: I think this is ok for now.

bbengfort (approved):
Thanks for adding the validation code to the options! It might be a little tightly coupled right now but we're still figuring out the SDK so I think this is the right path forward.

PR #223, opened 2023-02-16
TopicDelete Ensign integration

pdeziel:
### Scope of changes

This updates the `TopicDelete` endpoint so that it actually calls Ensign to delete the topic. This endpoint is protected by a confirmation token to prevent accidental deletes, since this will destroy all of the data in the topic. The user must call the endpoint once to retrieve the confirmation token. At this point the frontend should prompt the user for a confirmation, then send a second request to the endpoint including the token to actually perform the delete.

Since the delete is actually performed in Ensign, the Tenant database object will stay around, although the frontend can interpret the `state` value on the topic in order to determine whether or not to render the resource for the user. In a future story we will actually implement a routine on Tenant which polls Ensign asynchronously and makes the database updates.

Fixes SC-13773

### Type of change

- [x] new feature
- [ ] bug fix
- [ ] documentation
- [ ] testing
- [ ] technical debt
- [ ] other (describe)

### Acceptance criteria

See checklist

### Author checklist

### Reviewer(s) checklist

- [ ] Does the confirmation flow make sense?


pkg/utils/secrets/token.go
```diff
@@ -0,0 +1,37 @@
...
+
+// ValidateToken checks if a token contains any invalid characters.
+func ValidateToken(token string) bool {
+	for _, c := range token {
+		if !strings.ContainsRune(string(chars), c) && !strings.ContainsRune(string(specialChars), c) {
+			return false
+		}
+	}
+	return true
+}
```
pdeziel, line 37: Copied directly from GDS

pkg/tenant/topics_test.go
```diff
@@ -481,30 +504,45 @@ func (suite *tenantTestSuite) TestTopicDelete() {
...
 	// Should return an error if the topic ID is parsed but not found.
-	trtl.OnDelete = func(ctx context.Context, dr *pb.DeleteRequest) (*pb.DeleteReply, error) {
+	trtl.OnGet = func(ctx context.Context, gr *pb.GetRequest) (*pb.GetReply, error) {
 		return nil, errors.New("key not found")
 	}
 
-	err = suite.client.TopicDelete(ctx, "01GNA926JCTKDH3VZBTJM8MAF6")
-	suite.requireError(err, http.StatusNotFound, "could not delete topic", "expected error when topic ID is not found")
+	_, err = suite.client.TopicDelete(ctx, req)
+	suite.requireError(err, http.StatusNotFound, "topic not found", "expected error when topic ID is not found")
```
pdeziel, line 547: The rest of the endpoint will be tested in SC-13774

pkg/tenant/api/v1/api.go
```diff
@@ -50,7 +50,7 @@ type TenantClient interface {
 	TopicList(context.Context, *PageQuery) (*TopicPage, error)
 	TopicDetail(ctx context.Context, id string) (*Topic, error)
 	TopicUpdate(context.Context, *Topic) (*Topic, error)
-	TopicDelete(ctx context.Context, id string) error
+	TopicDelete(ctx context.Context, in *Confirmation) (*Confirmation, error)
```
bbengfort, line 53: Well implemented; I like that the same endpoint is used in two different ways.

pkg/tenant/db/topics.go
```diff
@@ -13,12 +14,15 @@ import (
...
-	Name      string    `msgpack:"name"`
-	Created   time.Time `msgpack:"created"`
-	Modified  time.Time `msgpack:"modified"`
+	OrgID               ulid.ULID                `msgpack:"org_id"`
+	ProjectID           ulid.ULID                `msgpack:"project_id"`
+	ID                  ulid.ULID                `msgpack:"id"`
+	Name                string                   `msgpack:"name"`
+	State               pb.TopicTombstone_Status `msgpack:"state"`
+	ConfirmDeleteToken  string                   `msgpack:"confirm_delete_token"`
+	ConfirmDeleteExpire time.Time                `msgpack:"confirm_delete_expire"`
```
bbengfort, line 23: I'd recommend the token be a base64 encoded msgpack data structure that includes the topic ID and expiration time stamp as well as a cryptographically randomly generated string. That way the token is self composed and in the future we can do things like cryptographically sign the token for further security.

pkg/tenant/api/v1/api.go
```diff
@@ -66,6 +66,14 @@ type TenantClient interface {
 // Top Level Requests and Responses
 //===========================================================================
 
+// Confirmation allows APIs to protect users from unintended actions such as deleting
+// data by including a confirmation token in the request.
+type Confirmation struct {
+	ID           string `json:"id"`
+	Name         string `json:"name"`
```
bbengfort, line 73: Is the name something you're thinking the user should supply?

pkg/tenant/topics.go
```diff
@@ -257,30 +262,137 @@ func (s *Server) TopicUpdate(c *gin.Context) {
 	c.JSON(http.StatusOK, t.ToAPI())
 }
 
-// TopicDelete deletes a topic from a user's request with a given ID
-// and returns a 200 OK response instead of an error response.
+// TopicDelete completely destroys a topic, removing the metadata in Trtl and as well
+// as all of the data in Ensign. Because this is irreversible, the first call returns
+// a confirmation token to the user. The user must provide this token in a subsequent
+// request in order to confirm the deletion. Because this operation is asynchronous,
+// the endpoint returns a 202 Accepted response.
```
bbengfort, line 269: Good choice of status code!

pkg/tenant/topics.go
```diff
@@ -257,30 +262,137 @@ func (s *Server) TopicUpdate(c *gin.Context) {
...
+
 	// Get the topic ID from the URL and return a 400 response
-	// if the topic does not exist.
+	// if the ID is not parseable
 	var topicID ulid.ULID
 	if topicID, err = ulid.Parse(c.Param("topicID")); err != nil {
-		log.Error().Err(err).Msg("could not parse topic ulid")
-		c.JSON(http.StatusBadRequest, api.ErrorResponse("could not parse topic ulid"))
+		log.Warn().Err(err).Msg("could not parse topic id")
+		c.JSON(http.StatusBadRequest, api.ErrorResponse("could not parse topic id"))
```
bbengfort, line 298: I'm thinking I'd actually prefer to return a 404 here (I know this has nothing to do with your PR, but just thinking about it again. E.g. if the user puts in `/project/:projectID/topics/foo` - that's not parseable as a ULID, but it should be not found not 400 right? The 404 is a little more secure and will allow us to change ID types in the future. 

```suggestion
		c.JSON(http.StatusNotFound, api.ErrorResponse("topic not found"))
```

pkg/tenant/topics.go
```diff
@@ -257,30 +262,137 @@ func (s *Server) TopicUpdate(c *gin.Context) {
...
+	if confirm.ID != topicID.String() {
+		log.Warn().Err(err).Msg("topic id in request body does not match topic id in URL")
+		c.JSON(http.StatusBadRequest, api.ErrorResponse("id in request body does not match id in URL"))
+		return
+	}
+
+	// Fetch the topic metadata from the database
+	var topic *db.Topic
+	if topic, err = db.RetrieveTopic(ctx, topicID); err != nil {
+		log.Error().Err(err).Str("topicID", topicID.String()).Msg("could not retrieve topic")
```
bbengfort, line 320: Could we add a check for the error type here; e.g. if the error is `IsNotFound` - I'd prefer not to have a `log.Error()` trigger an alert in Slack but if it's a database error then we should be alerted.

pkg/tenant/topics.go
```diff
@@ -257,30 +262,137 @@ func (s *Server) TopicUpdate(c *gin.Context) {
...
+		c.JSON(http.StatusNotFound, api.ErrorResponse("topic not found"))
+		return
+	}
+
+	// Verify that the user owns the topic
+	if claims.OrgID != topic.OrgID.String() {
+		log.Warn().Err(err).Str("user_org", claims.OrgID).Str("topic_org", topic.OrgID.String()).Msg("topic OrgID does not match user OrgID")
+		c.JSON(http.StatusForbidden, api.ErrorResponse("user is not authorized to delete this topic"))
```
bbengfort, line 328: This is not for this PR - but in general I've been thinking about security principles with our responses. If we return a 403 here then we're letting someone know that the topic exists and belongs to a different organization -- which could be information they could use to get a topic ID to do other attacks. 

E.g. a user simply has to register for an account to get valid claims then submit API requests with different Topic IDs - once they get a 403 instead of a 404 they know they have a good topic id belonging to someone else. 

Because of this I think it is probably better to return a 404 here; e.g. semantically it is "no topic found belonging to that organization". But that gives no information that the topic even exists. 


```suggestion
		c.JSON(http.StatusNotFound, api.ErrorResponse("topic not found"))
```

pkg/tenant/topics.go
```diff
@@ -257,30 +262,137 @@ func (s *Server) TopicUpdate(c *gin.Context) {
...
+		confirm.Name = topic.Name
+		confirm.ConfirmToken = topic.ConfirmDeleteToken
+		c.JSON(http.StatusOK, confirm)
+		return
+	}
+
+	// Verify that the confirmation token is correct
+	if confirm.ConfirmToken != topic.ConfirmDeleteToken {
+		log.Warn().Err(err).Msg("confirmation token does not match")
+		c.JSON(http.StatusBadRequest, api.ErrorResponse("unrecognized confirmation token"))
```
bbengfort, line 352: I think we should return a 412 Precondition Failed here to let the user know they should try the delete without the confirmation token again and then use a valid confirmation token.

```suggestion
		c.JSON(http.StatusPreconditionFailed, api.ErrorResponse("unrecognized confirmation token"))
```

pkg/tenant/topics.go
```diff
@@ -257,30 +262,137 @@ func (s *Server) TopicUpdate(c *gin.Context) {
...
+		return
+	}
+
+	// Verify the confirmation token has not expired
+	if time.Now().After(topic.ConfirmDeleteExpire) {
+		log.Warn().Err(err).Msg("confirmation token has expired")
+		c.JSON(http.StatusBadRequest, api.ErrorResponse("confirmation token has expired"))
```
bbengfort, line 359:
```suggestion
		c.JSON(http.StatusPreconditionFailed, api.ErrorResponse("confirmation token has expired"))
```

bbengfort: Going to be fun to document this in Postman!

pkg/tenant/topics.go
```diff
@@ -257,30 +262,137 @@ func (s *Server) TopicUpdate(c *gin.Context) {
...
 
-	// Delete the topic and return a 404 response if it cannot be removed.
-	if err = db.DeleteTopic(c.Request.Context(), topicID); err != nil {
-		log.Error().Err(err).Str("topicID", topicID.String()).Msg("could not delete topic")
-		c.JSON(http.StatusNotFound, api.ErrorResponse("could not delete topic"))
+	// The delete request is asynchronous so just update the state in the database
+	topic.State = tombstone.State
+	if err = db.UpdateTopic(ctx, topic); err != nil {
+		log.Error().Err(err).Str("topicID", topicID.String()).Msg("could not update topic state")
+		c.JSON(http.StatusInternalServerError, api.ErrorResponse("could not delete topic"))
```
bbengfort, line 393: Nicely done!

pkg/utils/secrets/token.go
bbengfort: Doh, not sure why we didn't use `crypto/rand` in GDS - but that's probably a problem. Would you mind copying the  `quarterdeck/keygen` package instead?

https://github.com/rotationalio/ensign/blob/main/pkg/quarterdeck/keygen/keygen.go

(You can use it directly if you'd like, but copying it into a secrets utils is fine; I'd like them decoupled so that we can change the way that Quarterdeck issues api keys; e.g. either use the keygen package directly or port it to utils so quarterdeck does not have a dependency on utils). 

I'll create a story to fix GDS.

bbengfort (commented):
Great start! I had an alternative suggestion for the token; let me know what you think.

pkg/utils/secrets/token.go
pdeziel: Sure that works for me

pkg/tenant/api/v1/api.go
pdeziel: I think it would be part of the response, so the user shouldn't supply it in the request but it gives the frontend something to display in the confirmation prompt without having to reference the original object.

pkg/tenant/topics.go
pdeziel: Agreed, I think we've gone back and forth but it does seem to make a bit more sense to return 404 since the ID is a part of the URL.

pkg/tenant/topics.go
pdeziel: Yeah that makes sense, at some point we might need to audit the other Tenant endpoints to make sure we're being consistent about the responses and not accidentally exposing info.

bbengfort: Could you create a future story for that?

pkg/tenant/api/v1/api.go
```diff
@@ -69,9 +69,9 @@ type TenantClient interface {
 // Confirmation allows APIs to protect users from unintended actions such as deleting
 // data by including a confirmation token in the request.
 type Confirmation struct {
-	ID           string `json:"id"`
-	Name         string `json:"name"`
-	ConfirmToken string `json:"confirm_token,omitempty"`
+	ID    string `json:"id,omitempty"`
+	Name  string `json:"name,omitempty"`
+	Token string `json:"token,omitempty"`
```
bbengfort, line 74: I agree this is a bit nicer.

pkg/tenant/topics.go
```diff
@@ -317,46 +316,61 @@ func (s *Server) TopicDelete(c *gin.Context) {
...
 		return
 	}
 
 	// Send confirmation token if not provided
-	if confirm.ConfirmToken == "" {
+	if confirm.Token == "" {
 		// Create a short-lived confirmation token in the database
-		topic.ConfirmDeleteToken = secrets.CreateToken(16)
-		topic.ConfirmDeleteExpire = time.Now().Add(5 * time.Minute)
+		if topic.ConfirmDeleteToken, err = db.NewResourceToken(topic.OrgID); err != nil {
```
bbengfort, line 339: Why is the resource token ID the topic's OrgID? I was expecting it to be the topic ID?

pkg/tenant/topics.go
```diff
@@ -317,46 +316,61 @@ func (s *Server) TopicDelete(c *gin.Context) {
...
+		c.JSON(http.StatusPreconditionFailed, api.ErrorResponse("invalid confirmation token"))
 		return
 	}
 
-	// Verify the confirmation token has not expired
-	if time.Now().After(topic.ConfirmDeleteExpire) {
-		log.Warn().Err(err).Msg("confirmation token has expired")
-		c.JSON(http.StatusBadRequest, api.ErrorResponse("confirmation token has expired"))
+	// Verify that we have the right token
+	if token.ID.Compare(topic.ID) != 0 {
```
bbengfort, line 371: Am I missing something? It looks like the OrgID is put in but then it is compared to the topic ID. If I am not missing anything why aren't the tests catching this?

bbengfort (approved):
Thank you for making those changes; I was confused about one thing but then it should be ready to go!

pkg/tenant/topics.go
pdeziel: My bad, it should be the topic ID and the test coverage isn't complete yet so I didn't catch it.

bbengfort: No worries - that's what code reviews are for!

PR #235, opened 2023-02-20
Update topic ensign integration

pdeziel:
### Scope of changes

This updates the `TopicUpdate` endpoint to call ensign. Currently only two fields are supported for update. The name can be updated which just changes the metadata in Trtl. The state can be updated, but only to the `READONLY` state. This actually implements a type of deletion where the topic is archived and the data can't be modified (although I suppose the metadata can still change).

Fixes SC-13775

### Type of change

- [x] new feature
- [ ] bug fix
- [ ] documentation
- [ ] testing
- [ ] technical debt
- [ ] other (describe)

### Acceptance criteria

See checklist below

### Author checklist

### Reviewer(s) checklist

- [ ] If archived topics cannot be modified, should this endpoint be foolproof protected like Delete?
- [ ] Is it too difficult to use this endpoint? Should we just have a readonly flag that can be set on the request instead of the state?


pkg/tenant/topics.go
```diff
@@ -237,24 +254,100 @@ func (s *Server) TopicUpdate(c *gin.Context) {
...
 	// Verify the topic name exists and return a 400 response if it doesn't.
 	if topic.Name == "" {
 		c.JSON(http.StatusBadRequest, api.ErrorResponse("topic name is required"))
 		return
 	}
 
-	// Get the specified topic from the database and return a 404 response if
-	// it cannot be retrieved.
+	// Topic state is optional but can only be set to readonly.
+	if topic.State != "" && topic.State != pb.TopicTombstone_READONLY.String() {
```
pdeziel, line 271: Potentially this is too confusing to use and we should just have an archive or readonly flag in the request?

pkg/tenant/topics.go
```diff
@@ -237,24 +254,100 @@ func (s *Server) TopicUpdate(c *gin.Context) {
...
+
+	// Verify that the user owns the topic
+	if claims.OrgID != t.OrgID.String() {
+		log.Warn().Str("user_org", claims.OrgID).Str("topic_org", t.OrgID.String()).Msg("user does not own topic")
 		c.JSON(http.StatusNotFound, api.ErrorResponse("topic not found"))
 		return
 	}
 
+	// Only archive mode is available on update
+	if topic.State == pb.TopicTombstone_READONLY.String() {
```
pdeziel, line 305: If there's no path for unarchiving a topic, should we also protect this endpoint with the confirmation token?

pkg/tenant/topics.go
bbengfort: Hmm, good point. The action of archiving makes the topic readonly -- and there is currently no mechanism to "unarchive" a topic (although we could potentially add this in the future). However, the state of the topic can change in Ensign and that might need to be reported back to the user. One thing we have to be careful of is to ensure that Tenant and Ensign are synchronized so that user actions in Tenant are reflected globally across the system. 

One way we could decouple this is by making archiving a distinct action rather than an update; e.g. having an archive endpoint; e.g. `POST /v1/topics/:topicID/archive`. This makes the action a little bit stronger and reduces the logic required for determining what the user wants to do. 

If we don't want another endpoint - then the logic for `PUT` is  `t.State != READONLY && topic.State != t.State` where `t` is the topic retrieved from the database. The semantic for `PUT` is that the user has to supply the entire document even if there are fields that they don't want to change (unlike `PATCH` where users specify only the fields they want to change).

pkg/tenant/topics.go
```diff
@@ -237,24 +254,100 @@ func (s *Server) TopicUpdate(c *gin.Context) {
...
+			c.JSON(http.StatusNotFound, api.ErrorResponse("topic not found"))
+			return
+		}
 		log.Error().Err(err).Str("topicID", topicID.String()).Msg("could not retrieve topic")
+		c.JSON(http.StatusInternalServerError, api.ErrorResponse("could not update topic"))
+		return
+	}
+
+	// Ensure the new name is valid
+	t.Name = topic.Name
```
bbengfort, line 290: The name will only be updated in Tenant, it won't be updated in Ensign. That's probably fine for now but we should create a follow on story for updating the name in Tenant.

pkg/tenant/topics.go
```diff
@@ -237,24 +254,100 @@ func (s *Server) TopicUpdate(c *gin.Context) {
...
+
+		// Archive means a "soft" delete in Ensign (no data is destroyed)
+		updateRequest := &pb.TopicMod{
+			Id:        t.ID.String(),
+			Operation: pb.TopicMod_ARCHIVE,
+		}
+		var tombstone *pb.TopicTombstone
+		if tombstone, err = s.ensign.DeleteTopic(ensignContext, updateRequest); err != nil {
+			if status.Code(err) == codes.NotFound {
+				log.Warn().Err(err).Str("topicID", updateRequest.Id).Msg("topic not found in ensign")
```
bbengfort, line 337:
```suggestion
				log.Error().Err(err).Str("topicID", updateRequest.Id).Msg("topic not found in ensign even though it is tenant")
```

bbengfort: This seems like a serious condition right?

pkg/tenant/topics.go
bbengfort: There may be a path for unarchiving in the future; I'm not sure though.

bbengfort (approved):

pkg/tenant/topics.go
pdeziel: Right, although potentially it could just be a syncing issue where the topic was just created in Tenant but it hasn't made its way through Ensign yet.

PR #253, opened 2023-02-25
Only refresh tokens should be allowed to reauthenticate

bbengfort:
### Scope of changes

Adds another audience to refresh tokens that identifies them as refresh tokens and verifies them in the Refresh handler. 

The basic fix is as follows, the `aud` claim in a JWT token can have multiple items. Access tokens have the audience claim `[https://api.rotational.app]` but refresh tokens now have the audience claim `[https://api.rotational.app, https://auth.rotational.app/v1/refresh]`. This means that the refresh tokens go through two checks; one that the audience claim `https://api.rotational.app` is good and then that `https://auth.rotational.app/v1/refresh` is in the token. 

I've added a test that shows that valid access tokens are rejected since they don't have this audience. 

Fixes SC-13078

### Type of change

- [ ] new feature
- [ ] bug fix
- [ ] documentation
- [ ] testing
- [x] technical debt
- [ ] other (describe)

### Acceptance criteria

I'm worried about the security of this model. Should we add a check that requires the access token, possibly expired be in the Authorization header then match the IDs of the two claims before we refresh? 

I kind of feel like we should; particularly if we put the access token in a Cookie that is http only and secure. 

@masskoder @pdamodaran any thoughts on this? 

### Author checklist

### Reviewer(s) checklist

- [ ] Should we add another test ensuring that the new access/refresh token have the same claims, particularly the audience? Just to make sure that it doesn't accidentally change in the future?

### DIFF
--- a/pkg/quarterdeck/auth.go
+++ b/pkg/quarterdeck/auth.go
@@ -360,6 +360,13 @@ func (s *Server) Refresh(c *gin.Context) {
 		return
 	}
 
+	// verify that the token is indeed a refresh token
+	if !claims.VerifyAudience(s.tokens.RefreshAudience(), true) {
+		log.Debug().Msg("token does not contain refresh audience")
+		c.JSON(http.StatusForbidden, api.ErrorResponse("could not verify refresh token"))
+		return
+	}
+
 	// get the user from the database using the ID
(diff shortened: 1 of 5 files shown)


pdamodaran (approved):
Looks good @bbengfort - I just added one suggested change for an additional check in the test.

PR #258, opened 2023-02-28
Email verification DB fields

pdeziel:
### Scope of changes

This adds a migration to add the email verification fields to the user table in the DB and updates the SQL handling code to create/read/update the new fields.

Fixes SC-14153

### Type of change

- [x] new feature
- [ ] bug fix
- [ ] documentation
- [ ] testing
- [ ] technical debt
- [ ] other (describe)

### Acceptance criteria

### Author checklist

### Reviewer(s) checklist


pkg/quarterdeck/db/migrations/0003_email_verification.sql
```diff
@@ -0,0 +1,16 @@
...
+BEGIN;
+
+-- Whether or not the user has a verified email address.
+ALTER TABLE users ADD COLUMN email_verified BOOL DEFAULT 0;
+
+-- Expiration time for the token.
+ALTER TABLE users ADD COLUMN email_verification_expires TEXT DEFAULT NULL;
+
+-- Token provided by the user to verify their email address.
+ALTER TABLE users ADD COLUMN email_verification_token TEXT DEFAULT NULL;
```
pdeziel, line 11: I tried to add a UNIQUE constraint here because we will need to look up the user record by the token. However, it seems that it's not allowed for the ALTER TABLE statement. Realistically it shouldn't matter since it's unlikely that two hashes would be the same but I was wondering if we wanted the constraint anyway.

bbengfort: The UNIQUE constraint also makes lookups faster since there is an index, so we probably would want the constraint. When you say it's not allowed is the problem that you can't do

`ALTER TABLE users ADD COLUMN email_verification_token TEXT UNIQUE DEFAULT NULL`? 

Is it possible to do

`CREATE UNIQUE INDEX unique_email_verification_token ON users (email_verification_token);`?

(note that SQLite does allow nullable unique fields and multiple NULL values do not count towards uniqueness -- but this is a SQLite thing not a SQL thing -- https://stackoverflow.com/questions/22699409/sqlite-null-and-unique).

bbengfort (approved):
Looks good - thanks for updating the model!

pkg/quarterdeck/db/migrations/0003_email_verification.sql
pdeziel: Thanks, I wasn't able to apply the constraint on the token column but I was able to add a unique index.

PR #267, opened 2023-03-02
Tenant API response cleanup

pdeziel:
### Scope of changes

This makes a pass at standardizing some of the tenant API responses. In this PR:

1. Endpoints were updated to use the shared `orgIDFromContext` method which saves a bit of code and standardizes the error responses when trying to parse the orgID from the user's claims.
2. If the requested object is not in the user's organization, we now return 404 instead of 403. This is a security thing to avoid leaking information about objects in other organizations to the user. For example the user could create a request with a projectID that's not in their organization. Even though they wouldn't be able to retrieve it, they could still gain information about whether the projectID exists based on the response (and potentially attempt malicious things with it).

Fixes SC-14066

### Type of change

- [ ] new feature
- [ ] bug fix
- [ ] documentation
- [ ] testing
- [x] technical debt
- [ ] other (describe)

### Acceptance criteria

Please review that the responses make sense to you.

### Author checklist

### Reviewer(s) checklist


daniellemaxwell (approved):
Looks good!

PR #277, opened 2023-03-06
Move tenant stats endpoint

pdeziel:
### Scope of changes

This moves the tenant stats endpoint to `/v1/stats` to reduce contention with the other tenant endpoints, and updates the API response so that Tenant can better define the stat values for the frontend.

Fixes SC-14623

### Type of change

- [ ] new feature
- [ ] bug fix
- [ ] documentation
- [ ] testing
- [x] technical debt
- [ ] other (describe)

### Acceptance criteria

### Author checklist

### Reviewer(s) checklist


pkg/tenant/api/v1/api.go
```diff
@@ -277,9 +277,11 @@ type ContactInfo struct {
...
+// StatValue contains a value for a named statistic which is meant to support a variety
 // of statistics endpoints.
-type StatCount struct {
-	Name  string `json:"name"`
-	Count int64  `json:"count"`
+type StatValue struct {
+	Name    string `json:"name"`
+	Value   string `json:"value"`
+	Units   string `json:"units,omitempty"`
+	Percent string `json:"percent,omitempty"`
```
bbengfort, line 286: Is percent always an independent percentage; e.g. if this is "Storage Used" the value might be "516" and the units are "MiB" and the percent would be something like "23%" e.g. indicating that 23% of the issued capacity is being used? And if there isn't a "total" then percent would be omitted? 

Since we have "units" in the return, is there a reason value and percent aren't numeric (e.g. float64)?

pkg/tenant/api/v1/client.go
```diff
@@ -266,12 +266,12 @@ func (s *APIv1) TenantDelete(ctx context.Context, id string) (err error) {
...
 }
 
-func (s *APIv1) TenantStats(ctx context.Context, id string) (out []*StatCount, err error) {
+func (s *APIv1) TenantStats(ctx context.Context, id string) (out []*StatValue, err error) {
 	if id == "" {
 		return nil, ErrTenantIDRequired
 	}
 
-	path := fmt.Sprintf("/v1/tenant/%s/stats", id)
+	path := fmt.Sprintf("/v1/stats/tenant/%s", id)
```
bbengfort, line 274: I know that you and Cletus discussed this, but I didn't see in the story why swapping the root of the route would reduce contention. Would you mind giving me a quick run down of the discussion?

bbengfort (approved):
I had some questions in the code - everything looks ok to me; but I'm not sure I totally understand the reasoning behind some of the choices.

pkg/tenant/api/v1/api.go
pdeziel: The idea is that Name and Value are always required (used for both counts and unit values). The Percent field operates as you described; it only exists for fields which have a "total" (although should we include the total as well?)

I suppose it probably makes more sense to have the value and percent fields be numeric; I was thinking that it didn't make sense for the "count" values to be floats but it might be simpler to have the frontend handle the string formatting depending on the Units field.

pkg/tenant/api/v1/client.go
pdeziel: I don't know the exact issue but the frontend was having trouble distinguishing the `/v1/tenant/:tenantID/projects` and  `/v1/tenant/:tenantID/stats` patterns. My suggestion was to move the stats endpoints to a different hierarchy since we might have more of them in the future.

pkg/tenant/api/v1/api.go
bbengfort: I agree; I can imagine the situation where they want to round to 2 decimal places or to none in different contexts, so numeric is better. Having a count as a float64 isn't accurate at longer precisions I don't think this endpoint needs extreme precision at all. 

If we have the percent as a float it is possible to compute the total from the value and the percent, so I don't think it necessarily needs to be included at this stage.

pkg/tenant/api/v1/client.go
bbengfort: hmm, ok - but likely we won't be willing to do that for other things in the object hierarchy so they will have to manage those routes at some point. Personally, I think the prefixes are easier because you can store the prefix and build all follow-on routes off them easily. With the reversal for `stats` that's no longer true.

bbengfort:
@pdeziel thank you for answering my questions; go ahead and merge when you're ready!

PR #280, opened 2023-03-07
API key permissions endpoints

pdeziel:
### Scope of changes

This adds endpoints to Tenant and Quarterdeck to retrieve the API key permissions available to the user. This allows the frontend to determine which permissions the user can set when creating API keys.

Fixes SC-14651

### Type of change

- [x] new feature
- [ ] bug fix
- [ ] documentation
- [ ] testing
- [ ] technical debt
- [ ] other (describe)

### Acceptance criteria

See checklist below

### Author checklist

### Reviewer(s) checklist

- [ ] Should we create shared functionality between QD and Tenant to avoid the passthrough call?
- [ ] Should we include "descriptions" in the response (loaded from the database or map in the code) or rely on directing users to documentation?

### DIFF
--- a/pkg/quarterdeck/apikeys.go
+++ b/pkg/quarterdeck/apikeys.go
@@ -12,6 +12,7 @@ import (
 	"github.com/rotationalio/ensign/pkg/quarterdeck/keygen"
 	"github.com/rotationalio/ensign/pkg/quarterdeck/middleware"
 	"github.com/rotationalio/ensign/pkg/quarterdeck/passwd"
+	perms "github.com/rotationalio/ensign/pkg/quarterdeck/permissions"
 	"github.com/rotationalio/ensign/pkg/quarterdeck/tokens"
 	"github.com/rotationalio/ensign/pkg/utils/pagination"
 	"github.com/rotationalio/ensign/pkg/utils/ulids"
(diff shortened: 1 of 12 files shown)


pkg/quarterdeck/apikeys.go
```diff
@@ -414,3 +415,33 @@ func (s *Server) APIKeyDelete(c *gin.Context) {
...
+		c.Error(err)
+		c.JSON(http.StatusUnauthorized, api.ErrorResponse("user claims unavailable"))
+		return
+	}
+
+	// All users can create API keys with publisher and subscriber permissions.
+	out := []string{
+		perms.Publisher,
+		perms.Subscriber,
+	}
```
pdeziel, line 437: One thing I was considering was adding descriptions to the response, although that would require a database lookup or duplicating them in code.

bbengfort: I think a database lookup then filtering based on the user's permissions makes sense -- it's a bit more flexible if we want to add permissions in the future than having a hard coded list here and this method may include permissions that are not eligible for an API key.

In terms of performance; either the front-end or tenant can cache the responses, so it shouldn't overwhelm quarterdeck.

pkg/quarterdeck/apikeys.go
```diff
@@ -414,3 +416,61 @@ func (s *Server) APIKeyDelete(c *gin.Context) {
...
+			c.JSON(http.StatusInternalServerError, api.ErrorResponse("could not fetch apikey permissions"))
+			return
+		}
+		out = append(out, permission)
+	}
+
+	// Filter other permissions based on the user's claims.
+	outf := make([]string, 0, len(out))
+	for _, permission := range out {
+		// TODO: we'll need a better way to identify permissions that both the user and the API key can have.
```
bbengfort, line 465: Can we create a future story for this TODO please?

pkg/quarterdeck/apikeys.go
bbengfort: I've added the database query, if you'd like to create a future story for adding descriptions, please feel free (might be a good user experience for the front-end).

bbengfort (approved):
Thank you so much for doing all the hard work of adding two API client methods and adding this endpoint! It will definitely enhance development on the front-end.

PR #320, opened 2023-03-16
Async retry tasks

pdeziel:
### Scope of changes

This modifies the task manager to support retry tasks.

Fixes SC-14225

### Type of change

- [x] new feature
- [ ] bug fix
- [ ] documentation
- [ ] testing
- [ ] technical debt
- [ ] other (describe)

### Acceptance criteria

See checklist below

### Author checklist

### Reviewer(s) checklist

- [ ] Do the retry and backoff configurations make sense for the current async tasks?
- [ ] Is it okay that the backoff logic blocks the worker thread?


pkg/utils/tasks/tasks.go
```diff
@@ -72,13 +110,64 @@ func (tm *TaskManager) IsStopped() bool {
...
+			// Note: This blocks the worker thread so queue sizes should be large
+			// enough to avoid blocking new tasks from being queued.
+			wait := time.After(conf.Backoff.NextBackOff())
+
+			select {
+			case <-ctx.Done():
+				err = ctx.Err()
+				break retryLoop
+			case <-wait:
+			}
```
pdeziel, line 145: This will block the worker thread, are we okay with this?

pkg/utils/tasks/tasks.go
```diff
@@ -34,6 +40,38 @@ type TaskManager struct {
 	stopped bool
 }
 
+// Option allows retries and backoff to be configured for individual tasks.
+type Option func(*options)
+
+type options struct {
+	Retries int
+	Backoff backoff.BackOff
```
bbengfort, line 48: These can be unexported.

pkg/utils/tasks/tasks.go
```diff
@@ -34,6 +40,38 @@ type TaskManager struct {
...
+		o.Retries = retries
+	}
+}
+
+// Backoff strategy to use when retrying, default is no backoff
+func WithBackoff(backoff backoff.BackOff) Option {
+	return func(o *options) {
+		o.Backoff = backoff
+	}
+}
```
bbengfort, line 65: Can we make the default exponential backoff -- this is the one we're going to use most I think. If we need no backoff in the future; we can make a struct that implements `backoff.Backoff` that has no timeout (if it doesn't already exist in the cenkalti package, and it might).

pkg/utils/tasks/tasks.go
```diff
@@ -34,6 +40,38 @@ type TaskManager struct {
...
+
+// Log an error if all retries failed under the provided context
```
bbengfort, line 67: By default does it simply log the error returned by the Task?

pkg/utils/tasks/tasks.go
pdeziel: Well this is easier said than done...I've been trying to come up with a solution where retry tasks can be added to the queue but still allow the `Stop()` to wait for the workers to exit. What makes it hard is that the channel can't be closed while retry tasks are still being added.

bbengfort: happy to talk this through if you'd like

pkg/utils/tasks/tasks.go
```diff
@@ -71,24 +121,120 @@ func (tm *TaskManager) IsStopped() bool {
...
+			return nil
+		}
+
+		// Log the error so we know that all the retries failed
+		if handler.conf.err != nil {
+			log.Error().Err(err).Int("attempts", handler.attempts).Msg("task failed after retries")
+
+			if handler.conf.ctx != nil {
+				// TODO: is this a thread-safe way to create a hub for capturing exceptions?
+				hub := sentrygin.GetHubFromContext(handler.conf.ctx).Clone()
```
bbengfort, line 147: I think this is a good idea, but there are a couple of things that we'll need to do to make it work:

1. The request context is going to go away in a different thread - so we need to clone before the task is queued.
2. We need a task context so the outer context can get passed in; right now no context will have the hub because it's being newly created and canceled inside of each task function. 

I'll go ahead and make that change to see if it fixes it.

pkg/utils/sentry/logger.go
```diff
@@ -119,6 +119,12 @@ func (e *Event) Str(key, value string) *Event {
 	return e
 }
 
+func (e *Event) Int(key string, value int) *Event {
+	e.extra[key] = value
+	e.zero = e.zero.Int(key, value)
+	return e
+}
```
bbengfort, line 126: :-)

pkg/utils/tasks/tasks.go
```diff
@@ -31,22 +40,61 @@ type TaskManager struct {
...
 
+// Option allows retries and backoff to be configured for individual tasks.
+type Option func(*options)
+
+type options struct {
+	retries int
+	backoff backoff.BackOff
+	ctx     *gin.Context
+	err     error
+}
```
bbengfort, line 56: I'm going to move this to its own file since I had a hard time locating it.

bbengfort (commented):
Looks great! Glad the scheduler fixed things. 

I have a couple of minor changes that I'm going to make right now so we can get it merged in.

pkg/utils/tasks/tasks.go
```diff
@@ -71,24 +121,120 @@ func (tm *TaskManager) IsStopped() bool {
...
+			for _, task := range tasks {
+				queue <- task
+			}
+			close(queue)
+			return
+		}
+	}
+}
+
+func TaskWorker(wg *sync.WaitGroup, queue <-chan *TaskHandler, retry <-chan *TaskHandler) {
```
bbengfort, line 226: Why is the task worker getting passed the `retry` channel?

bbengfort:
Well I discovered an issue that causes a memory leak. 

So if the queuing pattern is:

user -> queue -> scheduler -> tasks -> worker -> scheduler 

Then the `tasks` array can grow in an unbounded fashion.

If the queuing pattern is:

user -> queue -> scheduler -> worker -> queue -> scheduler -> tasks

Then this problem is prevented; but we can reach a deadlock where the a new task can't be scheduled because all workers are trying to queue a retry but there are no available workers to take a job to create room in the queue. 

Ooft.

bbengfort (approved):
I ended up doing a little more work on that than I expected -- sorry about that! Please review and let me know if you have any questions.

pkg/utils/tasks/tasks.go
```diff
@@ -31,64 +35,233 @@ type TaskManager struct {
...
+
+		h.parent.queueTask(h)
+		return
+	}
+
+	// At this point we've exhausted all possible retries, so log the error.
+	h.err.Since(h.scheduled)
+	h.err.Log(log.Logger)
+	h.err.Capture(sentry.GetHubFromContext(h.ctx))
+}
```
bbengfort, line 183: I moved the closure you created into a method on Exec and added a little logging.

pkg/utils/tasks/tasks.go
```diff
@@ -31,64 +35,233 @@ type TaskManager struct {
...
+			}
+
+			// Compute the new capacity, shrinking it if necessary to prevent leaks.
+			newcap := cap(pending)
+			if i+64 < newcap {
+				newcap = i + 64
+			}
+
+			pending = pending[:i:newcap]
+			log.Trace().Int("pending_length", len(pending)).Int("pending_capacity", cap(pending)).Msg("async task scheduler memory usage")
```
bbengfort, line 249: You didn't have this problem in your code since you we're copying into a `remaining` slice; but copying is a fairly heavy weight operation so this code prevents the memory leak while reducing the number of allocations and does not perform a copy. I think this is important for code that would be running once a second.

bbengfort: On monday, I might have you do some profiling to make sure that we've prevented the memory leak - would you be interested? If so, could you create a story for us and a meeting to take on that task?

pdeziel: I put a meeting on the calendar, I could also do this morning if that time doesn't work.

pkg/utils/tasks/tasks_test.go
```diff
@@ -2,30 +2,240 @@ package tasks_test
...
+	// First 50 have a retry, second 50 do not.
+	for i, retryTask := range state {
+		if i < 50 {
+			tm.Queue(retryTask, tasks.WithRetries(2), tasks.WithBackoff(&backoff.ZeroBackOff{}))
+		} else {
+			tm.Queue(retryTask, tasks.WithBackoff(&backoff.ZeroBackOff{}))
+		}
+	}
+
+	// Wait for all tasks to be completed and stop the task manager.
```
pdeziel, line 224: Should we be waiting for the waitgroup here?

pkg/utils/tasks/tasks_test.go
```diff
@@ -2,30 +2,240 @@ package tasks_test
...
+		state = append(state, &ErroringTask{failUntil: 5, wg: &wg})
+	}
+
+	// Queue state tasks with a retry limit that will ensure they all fail
+	for _, retryTask := range state {
+		tm.Queue(retryTask, tasks.WithRetries(1), tasks.WithBackoff(&backoff.ZeroBackOff{}))
+	}
+
+	// Wait for all tasks to be completed and stop the task manager.
+	time.Sleep(500 * time.Millisecond)
```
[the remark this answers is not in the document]
bbengfort: The failing tests don't mark the wait group as done - so we'd have to implement some logic so they know that they're failing and won't be retried in order for the wait group to work. That would be certainly be better than the sleep; it's a trade off between a realistic implementation for tests and doing concurrency correctly.

PR #327, opened 2023-03-17
Added rate limiter to quarterdeck

pdamodaran:
### Scope of changes

Added IP based rate limiting middleware to quarterdeck to control the number of requests that can be submitted by a single IP address.  

There is an issue with a test that submits multiple requests that should be returning a 429 error but is returning 200 instead.

Feel free to suggest changes to the naming or other design patterns that don't follow convention.

Fixes SC-13314

### Type of change

- [x] new feature
- [ ] bug fix
- [ ] documentation
- [ ] testing
- [ ] technical debt
- [ ] other (describe)

### Acceptance criteria

### Author checklist

### Reviewer(s) checklist


pkg/quarterdeck/middleware/ratelimiter.go
```diff
@@ -0,0 +1,117 @@
...
+	return i
+}
+
+// AddIP creates a new rate limiter and adds it to the ips map,
+// using the IP address as the key
+func (i *IPRateLimiter) AddIP(ip string) *rate.Limiter {
+	var onlyOnce sync.Once
+
+	var limiter *rate.Limiter
+	onlyOnce.Do(func() {
```
pdamodaran, line 51: @bbengfort - I think this takes care of the double checked locking but let me know if this is not correct.

pkg/quarterdeck/middleware/ratelimiter_test.go
```diff
@@ -0,0 +1,91 @@
...
+	done := make(chan bool)
+
+	go func() {
+		for {
+			select {
+			case <-done:
+				return
+			case <-ticker.C:
+				router.ServeHTTP(w, req)
+				//fmt.Println(w.Code)
```
pdamodaran, line 79: I have print statements here and in the rate limiter code.  If you uncomment these and run this test, you will see that the test hits the section in the rate limiter code that sends the 429 error response but the code that actually gets written to the writer is 200.

pkg/quarterdeck/config/config.go
```diff
@@ -32,6 +32,7 @@ type Config struct {
 	Database      DatabaseConfig
 	Token         TokenConfig
 	Sentry        sentry.Config
+	RateLimit     RatelimitConfig
```
bbengfort, line 35:
```suggestion
	RateLimit     RatelimitConfig `split_words:"true"`
```

☝🏽 will change the environment variable from `$QUARTERDECK_RATELIMIT_BURST` to `$QUARTERDECK_RATE_LIMIT_BURST`.

pkg/quarterdeck/config/config.go
```diff
@@ -50,6 +51,16 @@ type TokenConfig struct {
...
 }
 
+// Used by the rate limiter
+// Limit: represents the number of tokens that can be added to the token bucket per second
+// Burst: maximum number of tokens/requests in a "token bucket" and is initially full
+// Ttl: //number of minutes before an IP is removed from the ratelimiter map
+type RatelimitConfig struct {
+	Limit float64       `default0:"10"`
+	Burst int           `default0:"120"`
+	Ttl   time.Duration `default0:"5"`
```
bbengfort, line 61: This might make configuration a little easier:

```suggestion
	PerSecond float64       `default:"20" split_words:"true"`
	Burst int           `default:"120"`
	TTL   time.Duration `default:"5"`
```

Would you also mind updating the tests with the new config variables?

pkg/quarterdeck/middleware/ratelimiter.go
bbengfort: It's possible that I'm misinterpreting this but it seems that the `GetLimiter` function grabs an Rlock, then unlock and calls AddIP; the `AddIP` function creates `onlyOnce` inside of the function; so that mutex is locally scoped. Inside of the function, `onlyOnce` will only be executed once, however multiple concurrent calls to `AddIP` will each have their own `onlyOnce` and will not be synchronized; so it will be overwrritten if the calls happen in quick succession (which may be what is happening in the tests). 

Double checked locking would be:

```
func (i *IPRateLimiter) AddIP(ip string) (*rate.Limiter) {
	i.mu.Lock()
	defer i.mu.Unlock()

	// Here is the second check, e.g. the double check
	if ipInfo, exists := i.ips[ip]; exists {
		return ipInfo.limiter
	}

	// Otherwise the condition from the RLock is still true, so create the limiter 
	limiter := rate.NewLimiter(i.limit, i.burst)
	i.ips[ip] = &ipInfo{limiter, time.Now()}
	return limiter 
}
```

pkg/quarterdeck/middleware/ratelimiter.go
```diff
@@ -0,0 +1,117 @@
...
+
+	if !exists {
+		i.mu.RUnlock()
+		return i.AddIP(ip)
+	}
+
+	i.mu.RUnlock()
+
+	// update the last seen time associated with the IP address
+	ipInfo.lastSeen = time.Now()
```
bbengfort, line 75: A concurrency guard on last seen might be required; e.g. a mutex in ipInfo.

pkg/quarterdeck/middleware/ratelimiter.go
```diff
@@ -0,0 +1,117 @@
...
+			if time.Since(ipInfo.lastSeen) > ttl*time.Minute {
+				deleteList = append(deleteList, ip)
+			}
+		}
+		var mu sync.RWMutex
+		for _, ip := range deleteList {
+			mu.Lock()
+			delete(i.ips, ip)
+			mu.Unlock()
+		}
```
bbengfort, line 95: Make sure you're using the shared mutex!

```suggestion
		for _, ip := range deleteList {
			i.mu.Lock()
			delete(i.ips, ip)
			i.mu.Unlock()
		}
```

pkg/quarterdeck/middleware/ratelimiter.go
```diff
@@ -0,0 +1,117 @@
...
+	var limiter = NewIPRateLimiter(rate.Limit(conf.Limit), conf.Burst)
+	//run `cleanupIPInfo` in a go routine to periodically remove entries from the map
+	go limiter.cleanupIPInfo(conf.Ttl)
+	return func(c *gin.Context) {
+		limiter := limiter.GetLimiter(c.Request.RemoteAddr)
+		if !limiter.Allow() {
+			c.Writer.Header().Add("Ratelimit-Request-Remote-Addr", c.Request.RemoteAddr)
+			c.Writer.Header().Add("Ratelimit-Remaining", fmt.Sprintf("%.2f", limiter.Tokens()))
+			//fmt.Println(http.StatusTooManyRequests)
+			c.AbortWithStatusJSON(http.StatusTooManyRequests, api.ErrorResponse(ErrRateLimit))
```
bbengfort, line 109: The abort should cancel all processing so I'm not 100% sure what is going on with the tests. I'll have to play around with it a little bit.

pkg/quarterdeck/middleware/ratelimiter.go
pdamodaran: understood - so that point behind the double check is to do the check in the `AddIP` method to ensure that ipInfo does not exist before adding the IP address to the map.

pkg/quarterdeck/middleware/ratelimiter.go
```diff
@@ -0,0 +1,117 @@
...
+
+func RateLimiter(conf config.RatelimitConfig) gin.HandlerFunc {
+	var limiter = NewIPRateLimiter(rate.Limit(conf.Limit), conf.Burst)
+	//run `cleanupIPInfo` in a go routine to periodically remove entries from the map
+	go limiter.cleanupIPInfo(conf.Ttl)
+	return func(c *gin.Context) {
+		limiter := limiter.GetLimiter(c.Request.RemoteAddr)
+		if !limiter.Allow() {
+			c.Writer.Header().Add("Ratelimit-Request-Remote-Addr", c.Request.RemoteAddr)
+			c.Writer.Header().Add("Ratelimit-Remaining", fmt.Sprintf("%.2f", limiter.Tokens()))
```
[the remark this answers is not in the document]
pdamodaran: I have made the suggested changes - let me know if you would like for me to add `X-RateLimit-Request-Remote-Addr` as well.

pkg/quarterdeck/config/config.go
```diff
@@ -50,6 +51,17 @@ type TokenConfig struct {
 	RefreshOverlap  time.Duration     `split_words:"true" default:"-15m"`     // $QUARTERDECK_TOKEN_REFRESH_OVERLAP
 }
 
+// Used by the rate limiter
+// Limit: represents the number of tokens that can be added to the token bucket per second
+// Burst: maximum number of tokens/requests in a "token bucket" and is initially full
+// NOTE: if Burst is not included the config, then ALL REQUESTS WILL BE REJECTED!
+// Ttl: //number of minutes before an IP is removed from the ratelimiter map
+type RatelimitConfig struct {
```
pdamodaran, line 59: @bbengfort - if we don't include a value for `Burst` in the config, it defaults to 0 which means that every single request will get rejected.  Our tests will catch this, but should we set the defaults in `rate_limiter.go` if the values are not set?

pdamodaran:
@bbengfort - I saw the following error in the go 1.20 tests:

Screenshot 2023-03-21 at 7 57 55 AM

I added an assertion that after the third request, the rate limit remaining would equal 0.00 but in this case the rate limit remaining was 0.01 which is possible based on the calculations done in the https://pkg.go.dev/golang.org/x/time@v0.3.0/rate code.

I removed the assertion to see if the next test will fail if we run into a scenario where the rate limit remaining is 0.01 but I have run the test a couple of times and it has passed each time.  

In fact, before I made that change, I ran the failed test a second time, and it succeeded.

bbengfort:
Yeh, timing in tests is hard with Go; we have a couple of tests that fail occasionally in the code because of timing issues. Usually we just try to make things flexible; e.g. instead of an == 0 assertion, maybe a <= 0.1 assertion; etc. 

> @bbengfort - I saw the following error in the go 1.20 tests:
> 
> Screenshot 2023-03-21 at 7 57 55 AM
> 
> I added an assertion that after the third request, the rate limit remaining would equal 0.00 but in this case the rate limit remaining was 0.01 which is possible based on the calculations done in the https://pkg.go.dev/golang.org/x/time@v0.3.0/rate code.
> 
> I removed the assertion to see if the next test will fail if we run into a scenario where the rate limit remaining is 0.01 but I have run the test a couple of times and it has passed each time.
> 
> In fact, before I made that change, I ran the failed test a second time, and it succeeded.

pkg/quarterdeck/middleware/ratelimiter.go
```diff
@@ -0,0 +1,123 @@
...
+IPRateLimiter is an IP address based limiter that controls how frequently requests
+can be made from a single IP address.
+limit: represents the number of tokens that can be added to the token bucket per second
+burst: maximum number of tokens/requests in a "token bucket" and is initially full
+each request consumes tokens from the token bucket and if the bucket is empty
+when the request is made, the request is rejected
+*/
+type IPRateLimiter struct {
+	ips   map[string]*ipInfo //contains a map of IP address to information about its request pattern
+	mu    *sync.RWMutex
```
bbengfort, line 25: Rather than make this a `*sync.RWMutex` I'm going to embed it into the struct as a value; this makes it so the mutex can't be copied, which is the source of some subtle bugs.

pkg/quarterdeck/middleware/ratelimiter.go
```diff
@@ -0,0 +1,123 @@
...
+	go limiter.cleanupIPInfo(conf.TTL)
+	return func(c *gin.Context) {
+		// c.ClientIP() does a more thorough check to return the real client IP
+		// it also strips out the port, which ensures that we don't create multiple
+		// limiters for requests coming from the same IP address due to different port values
+		lim := limiter.GetLimiter(c.ClientIP())
+		if !lim.Allow() {
+			c.Writer.Header().Add("X-RateLimit-Limit", fmt.Sprintf("%v", lim.Burst()))
+			c.Writer.Header().Add("X-RateLimit-Remaining", fmt.Sprintf("%.2f", lim.Tokens()))
+			c.Writer.Header().Add("X-RateLimit-Reset", "1")
```
bbengfort, line 114: I'm going to shorten this section up a little bit to use `c.Header` directly and to only call the headers once.

pkg/quarterdeck/middleware/ratelimiter_test.go
```diff
@@ -0,0 +1,89 @@
...
+		rep, err := client.Do(req)
+		require.NoError(t, err)
+		// the first two requests will be allowed and rate limit remaining will be greater than 0
+		if i < 2 {
+			require.Equal(t, http.StatusOK, rep.StatusCode)
+			require.NotEqual(t, "0.00", rep.Header.Get("X-RateLimit-Remaining"))
+			// the third request will be allowed but rate limit remaining will be equal to zero
+		} else if i == 2 {
+			require.Equal(t, http.StatusOK, rep.StatusCode)
+			// beyond the third request all requests will be rejected
```
bbengfort, line 82: I'll add the less than or equal check back here; I think it's wise to do that.

pkg/quarterdeck/quarterdeck_test.go
```diff
@@ -77,6 +77,11 @@ func (s *quarterdeckTestSuite) SetupSuite() {
 			RefreshDuration: 20 * time.Minute,
 			RefreshOverlap:  -10 * time.Minute,
 		},
+		RateLimit: config.RateLimitConfig{
+			PerSecond: 20.00,
+			Burst:     120,
+			TTL:       5 * time.Minute,
+		},
```
bbengfort, line 84: Good catch!

bbengfort (approved):
Looks great; nicely done @pdamodaran -- thanks for dealing with those concurrency issues and implementing the double checked locking. I think the cleanup routine is also very good. 

I'm going to add some configuration documentation for devops purposes to this PR and make some minor changes that I commented on below; then I'll go ahead and merge this PR!

bbengfort:
On more thought I've also changed the per second default to 10 and the burst to 30; I realized that it was per second and when I suggested 20, I was thinking per minute I think. 10 requests per second is 100 ms per request and we expect about 200ms latencies from the front-end including processing. Although I expect Quarterdeck to respond < 75ms; if the requests are coming in faster than 100ms it means someone is scraping and not coming from the front-end. 

One concern that I had was that the Tenant IP address might be rate-limited. We need to ensure that the `X-Forwarded-For` IP address is set and if not, then set by Tenant; otherwise all requests will be rate limited by a single IP. I'll create a story for this.

PR #355, opened 2023-03-24
Sc 14300/org verification

daniellemaxwell:
### Scope of changes

Adds a verification method to the tenant db to confirm resources are on the correct organization.

Fixes SC-14300

### Type of change

- [x] new feature
- [ ] bug fix
- [ ] documentation
- [ ] testing
- [ ] technical debt
- [ ] other (describe)

### Acceptance criteria

### Author checklist

### Reviewer(s) checklist


pkg/tenant/api/v1/api.go
```diff
@@ -174,6 +174,7 @@ type Organization struct {
 }
 
 type Tenant struct {
+	OrgID           string `json:"org_id"`
```
daniellemaxwell, line 177: Before updating all of the handlers, I thought it would be best to discuss if we want to add the `OrgID` to each resource on the client side.

bbengfort: The OrgID on the client-side will principally come from the claims; so I think the user doesn't need to additionally specify the orgID - just the ID of the resource should be sufficient.

pkg/tenant/db/organizations.go
```diff
@@ -0,0 +1,56 @@
...
+const OrganizationNamespace = "organizations"
+
+type Organization struct {
+	ID       ulid.ULID `msgpack:"id"`
+	Name     string    `msgpack:"name"`
+	Created  time.Time `msgpack:"created"`
+	Modified time.Time `msgpack:"modified"`
+}
+
+var _ Model = &Organization{}
```
bbengfort, line 20: We don't need a whole model for this; what we're going to store is:

- `key`: the serialized ULID of the resource 
- `value`: the serialized ULID of the organization

So instead what we need is basically two functions; a function to retrieve the orgID from the database using the resourceID: 

```go
func GetOrgIndex(resourceID ulid.ULID) (orgID ulid.ULID, err error) {
    var data []byte

    // TODO: add namespace, trtl handling etc. 
    if data, err = db.Get(resourceID[:], nil); err != nil {
        return orgID, err 
    }

    if err = orgID.UnmarshalBinary(data); err != nil {
        return orgID, err
    }
    return orgID, nil
}
```

And a function to store the index pair:

```go
func PutOrgIndex(resourceID, orgID ulid.ULID) error {
    // TODO: add namespace, etc.
    if err = db.Put(resourceID[:], orgID[:], nil); err != nil {
        return err
    }
    return nil
}
```

Then everytime we create a resource; we need to make sure that the above index is correctly saved.

pkg/tenant/db/organizations.go
```diff
@@ -0,0 +1,56 @@
...
+	if ulids.IsZero(modelOrgID) {
+		return false, ErrMissingID
+	}
+
+	if orgID.Compare(modelOrgID) == 0 {
+		return true, nil
+	} else {
+		return false, ErrOrgNotVerified
+	}
+}
```
bbengfort, line 56: What I was thinking we'd do here is take the orgID from the claims and the resourceID from the request (e.g. from the URL or the body of the request); fetch the orgID from the database using the resourceID then compare the claimsID and the orgID. 

```suggestion
func VerifyOrg(claimsOrgID ulid.ULID, resourceID ulid.ULID) (bool, error) {
	if ulids.IsZero(orgID) {
		return false, ErrMissingOrgID
	}

	if ulids.IsZero(resourceID) {
		return false, ErrMissingID
	}
	
	orgID, err := GetOrgIndex(resourceID)
	if err != nil {
	    return false, err 
	}

	if orgID.Compare(claimsOrgID) == 0 {
		return true, nil
	} else {
		return false, ErrOrgNotVerified
	}
}
```

pkg/tenant/db/organizations.go
daniellemaxwell: The `Get` and `Put` methods in the tenant db take in a context and a model. Should `getRequest` and a `putRequest` be used for `GetOrgIndex` and `PutOrgIndex` and then set the namespace that way or should this be handled differently? https://github.com/rotationalio/ensign/blob/main/pkg/tenant/db/db.go#L168

pkg/tenant/tenants_test.go
```diff
@@ -353,6 +362,23 @@ func (suite *tenantTestSuite) TestTenantDetail() {
 	}
 	_, err = suite.client.TenantDetail(ctx, req.ID)
 	suite.requireError(err, http.StatusNotFound, "tenant not found", "expected error when tenant does not exist")
+
+	// Test VerifyOrg method and pass the resource ID as a value in the database.
```
daniellemaxwell, line 366: Placing this test after all of the other checks seemed to be the only place where I could get this to work without getting a trtl error.

pkg/tenant/db/organizations.go
```diff
@@ -0,0 +1,72 @@
...
+		return ErrMissingID
+	}
+
+	if ulids.IsZero(orgID) {
+		return ErrMissingOrgID
+	}
+
+	if err := putRequest(ctx, OrganizationNamespace, resourceID[:], orgID[:]); err != nil {
+		return err
+	}
```
bbengfort, line 41: Can we create a story for the following TODO please?

```suggestion
	// TODO: before doing the put request perform an existence check to make sure the key is not being overwritten
	if err := putRequest(ctx, OrganizationNamespace, resourceID[:], orgID[:]); err != nil {
		return err
	}
```

pkg/tenant/db/organizations.go
```diff
@@ -0,0 +1,72 @@
...
+
+	orgID, err := GetOrgIndex(ctx, resourceID)
+	if err != nil {
+		return false, err
+	}
+
+	err = PutOrgIndex(ctx, resourceID, orgID)
+	if err != nil {
+		return false, err
+	}
```
bbengfort, line 66: No `put` is necessary here; verification is a read-only operation, nothing has to be written to the database in this step. Instead the `PutOrgIndex` has to happen when a resource is created. 

```suggestion
```

pkg/tenant/db/organizations.go
```diff
@@ -0,0 +1,72 @@
...
+	if orgID.Compare(claimsOrgID) == 0 {
+		return true, nil
+	} else {
+		return false, ErrOrgNotVerified
```
bbengfort, line 70: @pdeziel do you want an error here in addition to the false?

This is kind of a chicken and an egg situation; the `err` implies that verification failed; but in this case verification "succeeded" in the sense that we proved that the org is not verified. So really it depends on how you want to perform this check. E.g. if you do

```go
if verified, err := VerifyOrg(ctx, orgID, resourceID); err != nil || !verified {
	return ErrNotAuthorized
}
```

Then it implies that `false` and an error will always be returned (which makes the bool a bit redundant). The other option is:

```go
if verified, err = VerifyOrg(ctx, orgID, resourceID); err != nil {
    // There was a database error
   sentry.Error(ctx).Err(err).Msg("could not check verification")
   return ErrInternal
}

if !verified {
   // Not verified
   return ErrNotAuthorized
}
```

Then the bool and the error make sense together; but we shouldn't return an error in the `false` case.

pkg/tenant/db/organizations.go
```diff
@@ -0,0 +1,72 @@
...
+	}
```
bbengfort, line 71: No need for `else` when returning: 

```suggestion
	if orgID.Compare(claimsOrgID) == 0 {
		return true, nil
	} 
	return false, ErrOrgNotVerified
```

pkg/tenant/db/organizations.go
```diff
@@ -0,0 +1,72 @@
...
+
+	if err = orgID.UnmarshalBinary(data); err != nil {
+		return ulid.ULID{}, err
+	}
+	return orgID, nil
+}
+
+// Store the resourceID as a key and the orgID as a value in the database.
+func PutOrgIndex(ctx context.Context, resourceID, orgID ulid.ULID) error {
```
bbengfort, line 30: Nicely done!

pkg/tenant/db/organizations_test.go
```diff
@@ -0,0 +1,52 @@
...
+	"context"
+
+	"github.com/oklog/ulid/v2"
+	"github.com/rotationalio/ensign/pkg/tenant/db"
+	pb "github.com/trisacrypto/directory/pkg/trtl/pb/v1"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
+)
+
+func (s *dbTestSuite) TestVerifyOrg() {
```
bbengfort, line 13: Good tests and good coverage.

pkg/tenant/apikeys.go
```diff
@@ -57,25 +57,12 @@ func (s *Server) ProjectAPIKeyList(c *gin.Context) {
...
-	}
-
-	// User should not be able to list API keys in another organization
-	if orgID.Compare(project.OrgID) != 0 {
-		sentry.Warn(c).Str("user_org", orgID.String()).Str("project_org", project.OrgID.String()).Msg("user cannot list API keys in this project")
-		c.JSON(http.StatusNotFound, api.ErrorResponse("project not found"))
-		return
-	}
+	// Verify user is on the correct organization.
+	db.VerifyOrg(c, orgID, projectID)
```
bbengfort, line 61: Wow; this dramatically simplifies the code here; that's great! 

We will need to do something with what `db.VerifyOrg` returns here; e.g. if not verified we'll want to return not found as we were doing before; and that will be true for all calls to `db.VerifyOrg` -- so a little more work is necessary; but it depends on how @pdeziel responds before we can proceed.

bbengfort (commented):
@daniellemaxwell I did an initial pass; but there is a question for @pdeziel that we need to proceed.

pkg/tenant/db/organizations.go
pdeziel: I think the way I've been handling these situations is to just return the error since it reduces the redundancy and simplifies the interface. However, we can add an error code check if we want to do conditional error logging.

```golang
if err = VerifyOrg(ctx, orgID, resourceID); err != nil {
    if !errors.Is(ErrNotFound) {
        sentry.Error(ctx).Err(err).Msg("could not check verification")
        return err
    }
    return err
}
```

bbengfort: Ok perfect; I'd just change it to a `sentry.Warn` instead of a `sentry.Error` but that works for me.

pkg/tenant/db/organizations.go
bbengfort, line 71: @daniellemaxwell ok, based on @pdeziel's response; I'd go with this:

```suggestion
func VerifyOrg(ctx context.Context, claimsOrgID, resourceID ulid.ULID) error {
	if ulids.IsZero(claimsOrgID) {
		return ErrMissingOrgID
	}

	if ulids.IsZero(resourceID) {
		return ErrMissingID
	}

	orgID, err := GetOrgIndex(ctx, resourceID)
	if err != nil {
		return err
	}

	if orgID.Compare(claimsOrgID) != 0 {
		return ErrOrgNotVerified
	} 
	return nil
```

pkg/tenant/members.go
```diff
@@ -123,6 +123,12 @@ func (s *Server) MemberCreate(c *gin.Context) {
 		return
 	}
 
+	// Verify the role provided is valid.
```
daniellemaxwell, line 126: This check was included with `MemberRoleUpdate`,  so I added it here in `MemberCreate` and `MemberUpdate` while updating the handlers and tests.

pkg/tenant/db/organizations.go
```diff
@@ -0,0 +1,70 @@
...
+		return err
+	}
+
+	return nil
+}
+
+// VerifyOrg will check that resources are allocated to the correct organization.
+// The method will take in an orgID from the claims and will return true if the orgID
+// from the database is the same and an error if it is not.
+func VerifyOrg(ctx context.Context, claimsOrgID, resourceID ulid.ULID) error {
```
bbengfort, line 51: Thank you for making these changes according to @pdeziel's preferences!

pkg/tenant/db/projects.go
```diff
@@ -102,6 +102,11 @@ func CreateTenantProject(ctx context.Context, project *Project) (err error) {
 		return err
 	}
 
+	// Store the project tenant ID as a key and project org ID as a value in the database for org verification.
+	if err = PutOrgIndex(ctx, project.TenantID, project.OrgID); err != nil {
+		return err
+	}
```
bbengfort, line 108: This is `CreateTenantProject` right? Shouldn't this be:

```suggestion
	if err = PutOrgIndex(ctx, project.ID, project.OrgID); err != nil {
		return err
	}
```

? We're not creating the tenant, but rather the project, unless I'm misunderstanding something. If we're also creating the tenant, then we also need to associate the tenant with the organization and need a second `PutOrgIndex` call for that resource.

pkg/tenant/members.go
```diff
@@ -257,6 +279,12 @@ func (s *Server) MemberUpdate(c *gin.Context) {
...
+	if !perms.IsRole(member.Role) {
+		c.JSON(http.StatusBadRequest, api.ErrorResponse("unknown member role"))
+		return
+	}
```
bbengfort, line 286: Good add; thanks for thinking outside of the PR!

bbengfort (approved):
@daniellemaxwell this was a big PR, thank you so much for working so hard on it! I made a couple of changes and pushed them directly in some of the critical sections. I had one comment/question if you could take a look at that. 

Beyond that, it was tough to review the members, projects, tenants, and topics RPC calls as thoroughly -- there is a lot here, this was a BIG PR! :-). There may be some things we have to clean up later, but my quick glance through seemed like everything was all right.

Very good work!

pkg/tenant/auth_test.go
```diff
@@ -134,7 +134,10 @@ func (s *tenantTestSuite) TestRegister() {
 
 	// Test that a tenant, member, and project were created without error
 	s.StopTasks()
-	require.Equal(4, trtl.Calls[trtlmock.PutRPC], "expected 3 put calls to trtl for namespaces tenant, member, project, and object_keys")
+	require.Equal(7, trtl.Calls[trtlmock.PutRPC], "expected 7 Put calls to trtl for two puts for each tenant, member, and project (store and org index) and one for object_keys.")
+	require.Equal(0, trtl.Calls[trtlmock.GetRPC], "expected no gets on register")
+	require.Equal(0, trtl.Calls[trtlmock.DeleteRPC], "expected no deletes on register")
+	require.Equal(0, trtl.Calls[trtlmock.CursorRPC], "expected no cursors on register")
```
bbengfort, line 140: I also fixed this test that was failing in CI - adding the additional `PutOrgIndex` calls doubles the number of calls on registration.

PR #357, opened 2023-03-29
Quarterdeck user invitations

pdeziel:
### Scope of changes

This adds a user invitations table to the database to support inviting users.

Fixes SC-15401

### Type of change

- [x] new feature
- [ ] bug fix
- [ ] documentation
- [ ] testing
- [ ] technical debt
- [ ] other (describe)

### Acceptance criteria

### Author checklist

### Reviewer(s) checklist


pkg/quarterdeck/db/models/users.go
```diff
@@ -344,6 +360,107 @@ func (u *User) UpdateLastLogin(ctx context.Context) (err error) {
...
+
+// Create an invitation in the database from the user to the specified email address
+// and return the invitation token to send to the user.
+func (u *User) Invite(ctx context.Context, email string) (token string, err error) {
+	var (
+		invite *db.VerificationToken
+	)
+
+	// Create a token that expires in 7 days
+	if invite, err = db.NewVerificationToken(email); err != nil {
```
pdeziel, line 375: I'm not sure yet if we need to include the orgID in the token or we will just have Quarterdeck do the database lookup.

pkg/quarterdeck/config/config.go
```diff
@@ -27,6 +27,7 @@ type Config struct {
 	LogLevel      logger.LevelDecoder `split_words:"true" default:"info"`                          // $QUARTERDECK_LOG_LEVEL
 	ConsoleLog    bool                `split_words:"true" default:"false"`                         // $QUARTERDECK_CONSOLE_LOG
 	AllowOrigins  []string            `split_words:"true" default:"http://localhost:3000"`         // $QUARTERDECK_ALLOW_ORIGINS
+	InviteBaseURL string              `split_words:"true" default:"https://rotational.app/invite"` // $QUARTERDECK_INVITE_BASE_URL
```
bbengfort, line 30: Can we put this in its own configuration and separate them out it to components to make adding links easier in the future? Something like:

```go
type EmailURLsConfig struct {
    Base string `default:"https://rotational.app"`
    Verify string `default:"/verify"`
    Invite  string `deafult:"/invite"` 
}

func (c EmailURLsConfig) VerifyURL() *url.URL {
    u, _ := url.Parse(c.Base)
    return u.ResolveReference(url.URL{Path: c.Verify})
}

// etc
… 
```

bbengfort: This will also make it easier to put the params on and you can pass just the `EmailURLsConfig` to the email manager rather than the whole config (you may have to pass two configs, one for the email manager settings and one for the URLs -- but the second one could be an interface rather than a config object to make testing easier).

pkg/quarterdeck/db/migrations/0004_user_invitations.sql
```diff
@@ -0,0 +1,19 @@
...
+    token               TEXT NOT NULL UNIQUE,
+    secret              BLOB NOT NULL,
+    created_by          BLOB NOT NULL,
+    created             TEXT NOT NULL,
+    modified            TEXT NOT NULL,
+    FOREIGN KEY (organization_id) REFERENCES organizations (id) ON DELETE CASCADE,
+    FOREIGN KEY (created_by) REFERENCES users (id) ON DELETE CASCADE
+);
+
+CREATE UNIQUE INDEX unique_user_invitation_token ON user_invitations (token);
```
bbengfort, line 17: I don't think you need this `CREATE UNIQUE INDEX` if you have the `UNIQUE` in the table definition; I think the reason we had to do that last time was because of `ALTER TABLE`.

pkg/quarterdeck/db/models/users.go
bbengfort: Either is fine with me.

bbengfort (approved):
Looks good!

PR #359, opened 2023-03-30
Sc 15395/member content

daniellemaxwell:
### Scope of changes

Adds fields to the `Member` model that will be included on the `Teams` page. This also added a validation check for the member's email in the db.

A minor update which changes `projectID` parse from a sentry.Warn to log.Warn is also included.

Fixes SC-15395 & SC-15202

### Type of change

- [x] new feature
- [ ] bug fix
- [ ] documentation
- [ ] testing
- [ ] technical debt
- [ ] other (describe)

### Acceptance criteria

### Author checklist

### Reviewer(s) checklist


pkg/tenant/api/v1/api.go
```diff
@@ -187,11 +187,15 @@ type TenantPage struct {
...
-	Created  string `json:"created,omitempty"`
-	Modified string `json:"modified,omitempty"`
+	ID           string `json:"id" uri:"id"`
+	Email        string `json:"string"`
+	Name         string `json:"name"`
+	Role         string `json:"role"`
+	Status       string `json:"status"`
+	Created      string `json:"created,omitempty"`
+	Modified     string `json:"modified,omitempty"`
+	DateAdded    string `json:"date_added,omitempty"`
```
pdeziel, line 197: Is `DateAdded` the same as `Created`?

pkg/tenant/api/v1/api.go
```diff
@@ -187,11 +187,15 @@ type TenantPage struct {
...
+	LastActivity string `json:"last_activity,omitempty"`
```
pdeziel, line 198: Is `LastActivity` just `LastLogin`? I think that's a great field to add, but just wanted to make sure I understand how it's going to be used.

pkg/tenant/db/members.go
```diff
@@ -64,6 +72,10 @@ func (m *Member) Validate() error {
 		return ErrMissingOrgID
 	}
 
+	if m.Email == "" {
+		return ErrMissingMemberEmail
+	}
```
pdeziel, line 77: Good check!

pkg/tenant/db/members.go
```diff
@@ -17,16 +17,24 @@ import (
...
+	Modified     time.Time `msgpack:"modified"`
+	DateAdded    time.Time `msgpack:"date_added"`
+	LastActivity time.Time `msgpack:"last_activity"`
 }
 
+type MemberStatus string
+
 var _ Model = &Member{}
 
+const MemberConfirmed = "Confirmed"
```
pdeziel, line 36: So I guess the states will be `Pending` and `Confirmed`? Are there any other statuses I'm missing?

pkg/tenant/db/members.go
```diff
@@ -104,6 +120,11 @@ func CreateMember(ctx context.Context, member *Member) (err error) {
 
 	member.Created = time.Now()
 	member.Modified = member.Created
+	member.DateAdded = member.Created
```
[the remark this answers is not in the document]
pdeziel: Oh I see, so `DateAdded` means when the teammate accepts the invitation?

pkg/tenant/db/members.go
```diff
@@ -104,6 +120,11 @@ func CreateMember(ctx context.Context, member *Member) (err error) {
...
+	member.LastActivity = member.Created
+
+	// TODO: Add check to display correct user status.
+	member.Status = MemberConfirmed
```
pdeziel, line 127: Since `member` is required for this method, I think we should leave the status up to the caller and rely on validation in this method to make sure the status field is set before saving to the database.

pdeziel (commented):
Looks good so far, I had a few questions but nothing especially complicated. Mainly I think we just need to figure out what timestamps the frontend needs vs. what we need to be saving in the database.

pkg/tenant/api/v1/api.go
daniellemaxwell: Yes, during the meeting we had I confirmed that `LastActivity` is the same as `LastLogin`. The design uses `Last Activity` as a label, so I used the same wording here.

pkg/tenant/db/members.go
daniellemaxwell: Yes, those are the only 2 statuses. Although, it is possible that we could add more, like `Inactive` for example.

pkg/tenant/db/members.go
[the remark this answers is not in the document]
daniellemaxwell: Yes, that's correct.

pkg/tenant/db/members.go
daniellemaxwell: Makes sense. I'll add the change.

pkg/tenant/db/members.go
```diff
@@ -82,11 +92,15 @@ func (m *Member) Validate() error {
...
-		Role:     m.Role,
-		Created:  TimeToString(m.Created),
-		Modified: TimeToString(m.Modified),
+		ID:           m.ID.String(),
+		Email:        m.Email,
+		Name:         m.Name,
+		Role:         m.Role,
+		Status:       m.Status,
+		Created:      TimeToString(m.Created),
+		Modified:     TimeToString(m.Modified),
```
pdeziel, line 101: One thing we might consider is removing the `Created` and `Modified` from the API struct, maybe instead of `Created` we can have `InvitedAt` (since that's a bit more useful for the frontend). Basically what I'm trying to say is that `Created` and `Modified` are more database-oriented and if we're not displaying them in the teammates table then we don't have to return them.

pkg/tenant/db/members.go
```diff
@@ -185,6 +206,8 @@ func UpdateMember(ctx context.Context, member *Member) (err error) {
 		member.Created = member.Modified
 	}
 
+	member.LastActivity = member.Modified
```
pdeziel, line 209: If `LastActivity` means `LastLogin` then do we only want to update it when the member logs in?

pkg/tenant/db/members.go
```diff
@@ -102,8 +116,15 @@ func CreateMember(ctx context.Context, member *Member) (err error) {
 		return err
 	}
 
+	// Verify user status exists.
+	if member.Status == "" {
+		return ErrMissingMemberStatus
+	}
```
pdeziel, line 122: We could move this inside of the validate method as well.

pdeziel (approved):
Looks good, thanks for dealing with the test updates! I just had a few notes for a future cleanup story.

pkg/tenant/db/members.go
daniellemaxwell: Sure. I don't think `Modified` is used anywhere, but `Created` is currently used on the `/profile` page for `Date Created`.

PR #363, opened 2023-04-03
Sc 15454/role update

daniellemaxwell:
### Scope of changes

This adds an endpoint that will allow users with the correct permission to update a member's role. If a change role request is submitted that would leave an organization without an "owner", then an error should be returned.

Fixes SC-15454

### Type of change

- [x] new feature
- [ ] bug fix
- [ ] documentation
- [ ] testing
- [ ] technical debt
- [ ] other (describe)

### Acceptance criteria

### Author checklist

### Reviewer(s) checklist

### DIFF
--- a/pkg/tenant/members.go
+++ b/pkg/tenant/members.go
@@ -280,6 +280,8 @@ func (s *Server) MemberRoleUpdate(c *gin.Context) {
 		return
 	}
 
+	// TODO: Add org verification
+
 	// Bind the user request with JSON.
 	params := &api.UpdateMemberParams{}
 	if err = c.BindJSON(&params); err != nil {
@@ -308,20 +310,13 @@ func (s *Server) MemberRoleUpdate(c *gin.Context) {
 		return
 	}
 
-	// Update member role. Required fields included for member validation check.
-	dbMember := &db.Member{
-		OrgID:  member.OrgID,
-		ID:     member.ID,
-		Email:  member.Email,
-		Name:   member.Name,
-		Role:   params.Role,
-		Status: member.Status,
-	}
+	// Update member role.
+	member.Role = params.Role
 
-	// Verify at least one user with owner permission remains in the organization.
+	// TODO: Verify at least one user with owner permission remains in the organization.
 
 	// Update member in the database.
-	if err = db.UpdateMember(c.Request.Context(), dbMember); err != nil {
+	if err = db.UpdateMember(c.Request.Context(), member); err != nil {
 		if errors.Is(err, db.ErrNotFound) {
 			c.JSON(http.StatusNotFound, api.ErrorResponse("member not found"))
 			return
@@ -332,7 +327,7 @@ func (s *Server) MemberRoleUpdate(c *gin.Context) {
 		return
 	}
 
-	c.JSON(http.StatusOK, dbMember.ToAPI())
+	c.JSON(http.StatusOK, member.ToAPI())
 }
 
 // MemberDelete deletes a member from a user's request with a given
--- a/pkg/tenant/members.go
+++ b/pkg/tenant/members.go
@@ -297,9 +297,13 @@ func (s *Server) MemberRoleUpdate(c *gin.Context) {
 	}
 
 	// Verify the role provided is valid.
-	perms.IsRole(params.Role)
+	if !perms.IsRole(params.Role) {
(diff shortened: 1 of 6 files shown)


pkg/tenant/members.go
```diff
@@ -261,6 +262,99 @@ func (s *Server) MemberUpdate(c *gin.Context) {
...
+			return
+		}
+
+		sentry.Error(c).Err(err).Msg("could not update member in the database")
+		c.JSON(http.StatusInternalServerError, api.ErrorResponse("could not update member"))
+		return
+	}
+
+	// Get members from the database and set page size to return all members.
+	// TODO: Create list method that will not require pagination for this endpoint.
```
daniellemaxwell, line 333: I've created a story for this in the next sprint.

pkg/tenant/members.go
```diff
@@ -261,6 +262,99 @@ func (s *Server) MemberUpdate(c *gin.Context) {
...
+
+	// Get the member ID from the URL and return a 400 if the member ID is not a ULID.
+	var memberID ulid.ULID
+	if memberID, err = ulid.Parse(c.Param("memberID")); err != nil {
+		sentry.Warn(c).Err(err).Str("id", c.Param("memberID")).Msg("could not parse member id")
+		c.JSON(http.StatusNotFound, api.ErrorResponse("member not found"))
+		return
+	}
+
+	// TODO: Add org verification
```
daniellemaxwell, line 283: This will be added after PR #355 is approved.

pdeziel (commented):
Thanks for adding the route! I had a few comments to be addressed, let me know if you have any questions.

pkg/tenant/members.go
daniellemaxwell: `MemberList` will have pagination. A separate method, like what we previously had, will be created for `MemberRoleUpdate`,  `TenantStats` and `getOwner` since we don't really need to pass in a cursor for these read requests.

pkg/tenant/members.go
```diff
@@ -261,6 +262,99 @@ func (s *Server) MemberUpdate(c *gin.Context) {
...
+		return
+	}
+
+	// Verify member role exists.
+	if params.Role == "" {
+		c.JSON(http.StatusBadRequest, api.ErrorResponse("member role is required"))
+		return
+	}
+
+	// Retrieve member from the database.
```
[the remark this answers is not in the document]
daniellemaxwell: I didn't see a story for this. I can create one and then it can be determine who it should be assigned to.

pkg/tenant/members.go
pdeziel: Ah, that makes sense and seems like a good abstraction.

pkg/tenant/members.go
```diff
@@ -261,6 +262,105 @@ func (s *Server) MemberUpdate(c *gin.Context) {
...
+	}
+
+	// Verify member role exists.
+	if params.Role == "" {
+		c.JSON(http.StatusBadRequest, api.ErrorResponse("member role is required"))
+		return
+	}
+
+	// Verify the role provided is valid.
+	perms.IsRole(params.Role)
```
pdeziel, line 300: Sorry, should have mentioned that this method just returns a bool so we do need to check the result and return a `StatusBadRequest` if necessary.

pkg/tenant/members.go
```diff
@@ -261,6 +262,105 @@ func (s *Server) MemberUpdate(c *gin.Context) {
...
+	}
+
+	// Loop over dbMember and count the number of members whose role is Owner to verify that at least one Owner remains in the organization.
+	var count bool
+	for _, dbMember := range members {
+		if dbMember.Role == perms.RoleOwner {
+			count = true
+			break
+		}
+	}
```
pdeziel, line 319: It's possible we will want to put this in a helper method, since I think we will also want to check this on MemberDelete (to prevent the only Owner in an organization from being deleted). Can we put a TODO here to note that for now?

pkg/tenant/members.go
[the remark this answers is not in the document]
pdeziel: Perfect, thanks!

pdeziel (approved):
Looks good, I think we just need to add a check around the `perms.IsRole` but otherwise looks good to go, also a quick case to verify that would also be good.

PR #367, opened 2023-04-04
Quarterdeck send invite route

pdeziel:
### Scope of changes

This adds a route to Quarterdeck for sending email invites to new users. When the endpoint is invoked, a user invitation record is added to the database and an email is sent to the email address in the request with a verification token.

Fixes SC-15467

### Type of change

- [x] new feature
- [ ] bug fix
- [ ] documentation
- [ ] testing
- [ ] technical debt
- [ ] other (describe)

### Acceptance criteria

### Author checklist

### Reviewer(s) checklist


pkg/quarterdeck/db/models/users.go
```diff
@@ -361,32 +363,56 @@ func (u *User) UpdateLastLogin(ctx context.Context) (err error) {
...
+		}
+		userID = ulids.New()
+	} else if _, err := GetOrgUser(ctx, user.ID, u.orgID); err != nil {
+		if !errors.Is(err, ErrNotFound) {
+			return nil, err
+		}
+		userID = user.ID
+	} else {
+		return nil, ErrUserOrgExists
 	}
```
bbengfort, line 400: This was a little hard to read on the first pass; is this code also what you were expecting?

I'm actually not sure if the below code will work tbh -- but I am mostly trying to figure out how to interpret the if/elses. 

```suggestion
	} else if _, err := GetOrgUser(ctx, user.ID, u.orgID) {
		// Ensure the user hasn't already been added to the organization
		if err == nil {
			return nil, ErrUserOrgExists
		}
		
		if err != nil && !errors.Is(err, ErrNotFound) {
			return nil, err
		}
		
		// Assign the existing user's ID to the invitation
		userID = user.ID
	}
```

pkg/quarterdeck/db/models/users.go
```diff
@@ -361,32 +363,56 @@ func (u *User) UpdateLastLogin(ctx context.Context) (err error) {
...
+		return nil, err
+	}
+
+	// Get the invitee's ID if they already exist or create a new ID
+	var userID ulid.ULID
+	if user, err := GetUserEmail(ctx, email, ulid.ULID{}); err != nil {
+		if !errors.Is(err, ErrNotFound) {
+			return nil, err
+		}
+		userID = ulids.New()
```
bbengfort, line 392: Some comments might be helpful here; so if we try to fetch their email and `err == nil` then you're expecting us to go to the `else if` block? It is a bit tricky to read.

bbengfort (approved):
Looks good; mind having @eschmier take a look at the email copy?

pkg/quarterdeck/db/models/users.go
pdeziel: I attempted to break up the ifs a bit, let me know if that's clearer.

bbengfort: definitely better; thanks!

PR #373, opened 2023-04-05
Cookie-based Authentication

bbengfort:
### Scope of changes

Adds a handler function to set the access and refresh tokens as cookies in the request and updates the middleware to fetch authorization information from the cookies if it is not set in the headers. This will only work for TLS connections.

Fixes SC-14388

### Type of change

- [x] new feature
- [ ] bug fix
- [ ] documentation
- [ ] testing
- [x] technical debt
- [ ] other (describe)

### Acceptance criteria

### Author checklist

### Reviewer(s) checklist


pkg/quarterdeck/middleware/auth.go
```diff
@@ -117,21 +122,27 @@ func Authorize(permissions ...string) gin.HandlerFunc {
 }
 
 // GetAccessToken retrieves the bearer token from the authorization header and parses it
-// to return only the JWT access token component of the header. If the header is missing
-// or the token is not available, an error is returned.
-// TODO: switch to regular expression based parsing.
```
bbengfort, line 122: Also sorted this TODO with the regular expression on line 32.

pkg/quarterdeck/middleware/auth.go
```diff
@@ -156,7 +167,7 @@ func ContextFromRequest(c *gin.Context) (ctx context.Context, err error) {
 		return nil, ErrNoRequest
 	}
 
-	// Add access token to context
+	// Add access token to context (from either header or cookie using Authenticate middleware)
 	ctx = req.Context()
 	if token := c.GetString(ContextAccessToken); token != "" {
 		ctx = api.ContextWithToken(ctx, token)
```
bbengfort, line 173: I'm pretty sure this is where the access token is fetched to set the token in the header to Quarterdeck. @pdeziel can you confirm? If we're getting the access token from the context; then the `Authenticate` middleware is setting the access token context and it's using `GetAccessToken` to do that, which fetches from the cookies … so this should work for a cookie authentication forwarded to Quarterdeck.

pkg/quarterdeck/middleware/auth_test.go
```diff
@@ -136,10 +137,93 @@ func TestGetAccessToken(t *testing.T) {
...
+	require.Equal(t, token, tks)
+}
+
+func TestGetAccessTokenCookie(t *testing.T) {
+	// Test both the GetAccessToken Cookie and the SetAuthTokens handler
+	router := gin.New()
+
+	auth, err := authtest.NewServer()
+	require.NoError(t, err, "could not start authentication server")
+	defer auth.Close()
```
bbengfort, line 159: Unfortunately the only way to test cookies is to set up TLS servers … so this test is _much_ more heavyweight than the other non-cookie tests.

pkg/quarterdeck/tokens/expires.go
```diff
@@ -0,0 +1,33 @@
...
+	return claims.ExpiresAt.Time, nil
+}
+
+func NotBefore(tks string) (_ time.Time, err error) {
+	var claims *jwt.RegisteredClaims
+	if claims, err = ParseUnverified(tks); err != nil {
+		return time.Time{}, err
+	}
+	return claims.NotBefore.Time, nil
+}
```
bbengfort, line 33: Used to extract expiration and not before timestamps without having to have public keys to verify the token -- this allows us to set max age on the cookies with limited parsing.

pkg/tenant/auth.go
```diff
@@ -199,6 +206,13 @@ func (s *Server) Refresh(c *gin.Context) {
 		return
 	}
 
+	// Set the access and refresh tokens as cookies for the front-end
+	if err := middleware.SetAuthTokens(c, reply.AccessToken, reply.RefreshToken, s.conf.Auth.CookieDomain); err != nil {
+		sentry.Error(c).Err(err).Msg("could not set access and refresh token cookies")
+		c.JSON(http.StatusInternalServerError, api.ErrorResponse("could not set auth cookies"))
+		return
+	}
+
```
bbengfort, line 215: Set the cookies in Login and Refresh -- anywhere else?

bbengfort: This is just Tenant, which is the only thing that has a frontend; should we also set it in Quarterdeck for future proofing? Will we ever have a UI that goes directly to Quarterdeck?

pkg/quarterdeck/tokens/expires.go
```diff
@@ -0,0 +1,33 @@
+package tokens
+
+import (
+	"time"
+
+	jwt "github.com/golang-jwt/jwt/v4"
+)
+
+var tsparser = &jwt.Parser{SkipClaimsValidation: true}
```
pdamodaran, line 9:
```suggestion
// Used to extract expiration and not before timestamps without having to use public keys
var tsparser = &jwt.Parser{SkipClaimsValidation: true}
```

pdamodaran (approved):
@bbengfort - I just have a couple of suggested changes for adding comments but this is good to go.

PR #381, opened 2023-04-07
sc-15453 Change Role Modal

elysee15:
### Scope of changes
Fixes SC-15453

### Type of change

- [x] new feature
- [ ] bug fix
- [ ] documentation
- [ ] testing
- [ ] technical debt
- [ ] other (describe)

### Acceptance criteria
image

https://user-images.githubusercontent.com/54010836/230677346-96019061-0e5c-436a-a6a1-026e5a114b42.mov

### Author checklist

- [x] I have manually tested the change and/or added automation in the form of unit tests or integration tests
- [x]  I have updated the dependencies list
- [ ]  I have recompiled and included new protocol buffers to reflect changes I made
- [ ]  I have added new test fixtures as needed to support added tests
- [x]  Check this box if a reviewer can merge this pull request after approval (leave it unchecked if you want to do it yourself)
- [x]  I have moved the associated Shortcut story to "Ready for Review"

### Reviewer(s) checklist

- [ ] Any new user-facing content that has been added for this PR has been QA'ed to ensure correct grammar, spelling, and understandability.
- [ ] Are there any TODOs in this PR that should be turned into stories?

PR #384, opened 2023-04-10
Verifies org has more than 1 owner before updating role in database

daniellemaxwell:
### Scope of changes

Fixes a bug that changed a member's role from `Owner` to another option in the database if there was only one `Owner` in the organization.

This was noticed in a video included with PR #381 when the only member in the org's role changed from `Owner` to `Admin`, but could not switch back from `Admin` to `Owner`,

Fixes SC-15894

### Type of change

- [ ] new feature
- [x] bug fix
- [ ] documentation
- [ ] testing
- [ ] technical debt
- [ ] other (describe)

### Acceptance criteria

### Author checklist

### Reviewer(s) checklist

### DIFF
--- a/pkg/tenant/members.go
+++ b/pkg/tenant/members.go
@@ -386,20 +386,20 @@ func (s *Server) MemberRoleUpdate(c *gin.Context) {
 		c.JSON(http.StatusInternalServerError, api.ErrorResponse("member id does not match id in URL"))
 	}
 
-	// Get members from the database and set page size to return all members.
-	// TODO: Create helper method to check if an organization has at least one owner.
-	// TODO: Create list method that will not require pagination for this endpoint.
-	getAll := &pg.Cursor{StartIndex: "", EndIndex: "", PageSize: 100}
-	var members []*db.Member
-	if members, _, err = db.ListMembers(c.Request.Context(), orgID, getAll); err != nil {
-		sentry.Error(c).Err(err).Msg("could not list members")
-		c.JSON(http.StatusInternalServerError, api.ErrorResponse("could not update member role"))
-		return
-	}
-
-	// Loop over dbMember and break out of the loop if there are at least two members with the role Owner. If there is only
-	// one owner, they will not be able to change their role.
+	// If the member to be updated is an owner, loop over dbMember and break out of the loop if there are at least two owners.
+	// If member is the only owner, their role cannot be changed.
 	if member.Role == perms.RoleOwner {
+		// Get members from the database and set page size to return all members.
+		// TODO: Create helper method to check if an organization has at least one owner.
+		// TODO: Create list method that will not require pagination for this endpoint.
+		getAll := &pg.Cursor{StartIndex: "", EndIndex: "", PageSize: 100}
+		var members []*db.Member
+		if members, _, err = db.ListMembers(c.Request.Context(), orgID, getAll); err != nil {
+			sentry.Error(c).Err(err).Msg("could not list members")
+			c.JSON(http.StatusInternalServerError, api.ErrorResponse("could not update member role"))
+			return
+		}
+
 		count := 0
 		for _, dbMember := range members {
 			if dbMember.Role == perms.RoleOwner {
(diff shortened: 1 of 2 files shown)


pkg/tenant/members.go
```diff
@@ -377,16 +377,19 @@ func (s *Server) MemberRoleUpdate(c *gin.Context) {
...
-	// Loop over dbMember and count the number of members whose role is Owner to verify that at least one Owner remains in the organization.
-	var count bool
+	// Loop over dbMember and break out of the loop if there are at least two members with the role Owner. If there is only
+	// one owner, they will not be able to change their role.
+	count := 0
 	for _, dbMember := range members {
 		if dbMember.Role == perms.RoleOwner {
-			count = true
-			break
+			count++
```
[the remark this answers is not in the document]
pdeziel: We should only do this check if the member to be modified is an owner right? For example, it should still be possible to change a user who is a `Member` to an `Owner` even if there is only one `Owner`.

pkg/tenant/members.go
```diff
@@ -377,16 +377,19 @@ func (s *Server) MemberRoleUpdate(c *gin.Context) {
...
+			if count >= 2 {
+				break
+			}
 		}
 	}
 
-	if !count {
+	if count <= 1 {
```
pdeziel, line 392: Can we make this a switch statement? We may need to return a different error if `count=0` because now we're preventing that invariant.

```golang
switch count {
case 0:
    // There are no owners, which shouldn't happen so raise a sentry error here and return internal server error
case 1:
    // There is only one owner, which is a user problem so return a bad request error
}
```

pdeziel (changes requested):
Thanks for fixing this bug, I think we need to change a logic a bit though to catch an edge case.

daniellemaxwell:
@pdeziel Yes, we should only do the check if an owner's role will be changed.

pkg/tenant/members.go
```diff
@@ -411,6 +386,38 @@ func (s *Server) MemberRoleUpdate(c *gin.Context) {
...
+					break
+				}
+			}
+		}
+		switch count {
+		case 0:
+			sentry.Warn(c).Err(err).Msg("could not find any owners")
+			c.JSON(http.StatusInternalServerError, api.ErrorResponse("organization must have at least one owner"))
+		case 1:
+			c.JSON(http.StatusBadRequest, api.ErrorResponse("unable to change role of only owner"))
```
pdeziel, line 417:
```suggestion
			c.JSON(http.StatusBadRequest, api.ErrorResponse("organization must have at least one owner"))
```

pkg/tenant/members_test.go
```diff
@@ -698,11 +706,17 @@ func (suite *tenantTestSuite) TestMemberRoleUpdate() {
 
 	// Should return an error if org does not have an owner.
 	members[0].Role = perms.RoleMember
+	members[1].Role = perms.RoleAdmin
 	_, err = suite.client.MemberRoleUpdate(ctx, "01ARZ3NDEKTSV4RRFFQ69G5FAV", &api.UpdateMemberParams{Role: perms.RoleObserver})
-	suite.requireError(err, http.StatusBadRequest, "organization must have at least one owner", "expected error when org does not have an owner")
+	suite.requireError(err, http.StatusInternalServerError, "organization must have at least one owner", "expected error when org does not have an owner")
```
pdeziel, line 711:
```suggestion
	suite.requireError(err, http.StatusInternalServerError, "could not update member role", "expected error when org does not have an owner")
```

pkg/tenant/members_test.go
```diff
@@ -698,11 +706,17 @@ func (suite *tenantTestSuite) TestMemberRoleUpdate() {
...
 
-	// Set a member role in the database to owner.
+	// Set database to have one owner. Should return an error if org does not have an owner.
 	members[0].Role = perms.RoleOwner
+	_, err = suite.client.MemberRoleUpdate(ctx, "01ARZ3NDEKTSV4RRFFQ69G5FAV", &api.UpdateMemberParams{Role: perms.RoleObserver})
+	suite.requireError(err, http.StatusBadRequest, "unable to change role of only owner", "expected error when org does not have an owner")
```
pdeziel, line 716:
```suggestion
	suite.requireError(err, http.StatusBadRequest, "organization must have at least one owner", "expected error when org does not have an owner")
```

pdeziel (approved):
Looks great! I just had some error message updates, hope you don't mind.

PR #407, opened 2023-04-14
Add user role update endpoint to Quarterdeck

pdamodaran:
### Scope of changes

Added an endpoint to update a user's role in quarterdeck and updated tenant to call quarterdeck to update a user's role.  

TODO: update tenant members_test.go for edge cases.

Note that this PR does not include the changes to rename `Collaborators` to `Teammates`

Fixes SC-15731

### Type of change

- [x] new feature
- [ ] bug fix
- [ ] documentation
- [ ] testing
- [ ] technical debt
- [ ] other (describe)

### Acceptance criteria

### Author checklist

### Reviewer(s) checklist


pkg/quarterdeck/db/models/errors.go
```diff
@@ -18,9 +18,11 @@ var (
 	ErrMissingModelID      = errors.New("model does not have an ID assigned")
 	ErrMissingKeyMaterial  = errors.New("apikey model requires client id and secret")
 	ErrInvalidSecret       = errors.New("apikey secrets should be stored as argon2 derived keys")
-	ErrMissingInviteRole   = errors.New("missing role for user invite")
-	ErrInvalidInviteRole   = errors.New("invalid role for user invite")
-	ErrInvalidInviteEmail  = errors.New("invalid email for user invite")
+	ErrMissingRole         = errors.New("missing role")
+	ErrInvalidRole         = errors.New("invalid role")
+	ErrNoOwnerRole         = errors.New("organization is missing an owner")
+	ErrOwnerRoleConstraint = errors.New("organization must have at least one owner")
```
pdamodaran, line 24: Open to suggestions on a different name to capture this requirement

pkg/tenant/members_test.go
```diff
@@ -678,20 +691,20 @@ func (suite *tenantTestSuite) TestMemberRoleUpdate() {
 	_, err = suite.client.MemberRoleUpdate(ctx, "01ARZ3NDEKTSV4RRFFQ69G5FAV", &api.UpdateMemberParams{Role: "Viewer"})
 	suite.requireError(err, http.StatusBadRequest, "unknown member role", "expected error when member role is not valid")
 
-	// Should return an error if the member id in the database does not match the id in the URL.
-	_, err = suite.client.MemberRoleUpdate(ctx, "01GQ2XB2SCGY5RZJ1ZGYSEMNDE", &api.UpdateMemberParams{Role: perms.RoleObserver})
-	suite.requireError(err, http.StatusInternalServerError, "member id does not match id in URL", "expected error when member id does not match")
+	// // Should return an error if the member id in the database does not match the id in the URL.
```
pdamodaran, line 694: these edge cases will need to be updated

pkg/quarterdeck/db/models/errors.go
pdeziel: I think the name makes sense to me.

pkg/quarterdeck/quarterdeck.go
```diff
@@ -271,16 +272,19 @@ func (s *Server) Routes(router *gin.Engine) (err error) {
...
-		// Users Resource
+		// Users Resource - endpoint for Admin and Owner users of the organization
+		// to manager other users in their organization
 		users := v1.Group("/users", authenticate)
 		{
 			users.GET("/:id", middleware.Authorize(perms.ReadCollaborators), s.UserDetail)
 			users.PUT("/:id", middleware.Authorize(perms.EditCollaborators), s.UserUpdate)
 			users.GET("", middleware.Authorize(perms.ReadCollaborators), s.UserList)
 			users.DELETE("/:id", middleware.Authorize(perms.RemoveCollaborators), s.UserDelete)
+			users.POST("/:id", middleware.Authorize(perms.EditCollaborators), s.UserRoleUpdate)
```
pdeziel, line 283: I remember discussing that we should have the role update be a POST, but feel free to correct me if I'm remembering wrong.

pkg/quarterdeck/api/v1/api.go
```diff
@@ -358,13 +359,15 @@ type OpenIDConfiguration struct {
...
-	Email       string               `json:"email"`
-	LastLogin   string               `json:"last_login"`
-	OrgID       ulid.ULID            `json:"org_id"`
-	OrgRoles    map[ulid.ULID]string `json:"org_roles"`
-	Permissions []string             `json:"permissions"`
+	UserID    ulid.ULID `json:"user_id"`
+	Name      string    `json:"name"`
+	Email     string    `json:"email"`
+	Role      string    `json:"role"`
+	LastLogin time.Time `json:"last_login"`
```
bbengfort, line 366: I think I would keep orgID here. I can see why we would remove OrgRoles and Permissions.

pkg/quarterdeck/api/v1/api.go
```diff
@@ -41,6 +41,7 @@ type QuarterdeckClient interface {
 
 	// Users Resource
 	UserUpdate(context.Context, *User) (*User, error)
+	UserRoleUpdate(context.Context, string, *UpdateRoleRequest) (*User, error)
```
bbengfort, line 44: Little nit, If you're going to have a request structure, I prefer for the userID to be in that data structure, e.g.

```
UserRoleUpdate(context.Context, *UpdateRoleRequest) (*User, error)

type UpdateRoleRequest struct {
    ID ulid.ULID
    Role string
}
```

pkg/quarterdeck/api/v1/client.go
```diff
@@ -348,6 +348,21 @@ func (s *APIv1) UserUpdate(ctx context.Context, in *User) (out *User, err error)
 	return out, nil
 }
 
+func (s *APIv1) UserRoleUpdate(ctx context.Context, id string, in *UpdateRoleRequest) (out *User, err error) {
+	endpoint := fmt.Sprintf("/v1/users/%s", id)
```
bbengfort, line 352: Also RE: my above comment; having the role as a ULID means that we can guarantee the endpoint is created correctly. You would just need to do something like:

```
endpoint := fmt.Sprintf("/v1/users/%s", in.ID)
```

pkg/quarterdeck/db/models/errors.go
bbengfort: Me too!

pkg/quarterdeck/db/models/users.go
```diff
@@ -1035,6 +1035,74 @@ func (u *User) fetchRoles(tx *sql.Tx) (err error) {
 	return rows.Err()
 }
 
+const (
+	getUserOrgRoleSQL = "SELECT r.name FROM organization_users ur JOIN roles r ON ur.role_id=r.id WHERE user_id=:userID AND organization_id=:orgID"
+	getNumOwnersSQL   = "SELECT COUNT(*) FROM organization_users WHERE organization_id=:orgID and role_id IN (SELECT id from roles where name=:ownerRole)"
+	userRoleUpdateSQL = "UPDATE organization_users SET modified=:modified, role_id=(SELECT id FROM roles WHERE name=:role) WHERE user_id=:userID AND organization_id=:orgID"
```
bbengfort, line 1041: Good SQL! You could use strict equality with the SELECT since we know that only one owner role will be returned but using IN is probably a bit more future proof.

Regarding `getNumOwnersSQL` -- I guess I'll find out soon but just from reading the SQL it seems like you're going to check for _at least 2_ rather than _at least 1_ since this query does not exclude the user being modified.

pkg/quarterdeck/db/models/users.go
```diff
@@ -1035,6 +1035,74 @@ func (u *User) fetchRoles(tx *sql.Tx) (err error) {
...
+		return ErrInvalidRole
+	}
+
+	var tx *sql.Tx
+	if tx, err = db.BeginTx(ctx, nil); err != nil {
+		return err
+	}
+	defer tx.Rollback()
+
+	u.SetModified(time.Now())
```
bbengfort, line 1067: Thank you for ensuring the struct is updated even though you're not saving it!

pkg/quarterdeck/db/models/users.go
```diff
@@ -1035,6 +1035,74 @@ func (u *User) fetchRoles(tx *sql.Tx) (err error) {
...
+
+	// Get the user's current role
+	var currentRole string
+	if err = tx.QueryRow(getUserOrgRoleSQL, sql.Named("userID", u.ID), sql.Named("orgID", userOrg)).Scan(&currentRole); err != nil {
+		if errors.Is(err, sql.ErrNoRows) {
+			return ErrUserOrganization
+		}
+		return err
+	}
```
bbengfort, line 1076: This is fine for now, but note that SQL like this may not work when we move to Raft replication since this is relying on Go logic in the middle of the transaction.

pkg/quarterdeck/db/models/users.go
```diff
@@ -1035,6 +1035,74 @@ func (u *User) fetchRoles(tx *sql.Tx) (err error) {
...
+		if err = tx.QueryRow(getNumOwnersSQL, sql.Named("orgID", userOrg), sql.Named("ownerRole", perms.RoleOwner)).Scan(&numOwners); err != nil {
+			return err
+		}
+
+		switch numOwners {
+		case 0:
+			return ErrNoOwnerRole
+		case 1:
+			return ErrOwnerRoleConstraint
+		}
```
bbengfort, line 1091: Case 0 shouldn't happen really - if it does, the database has been corrupted; so hopefully we'll see `ErrNoOwnerRole` in Sentry and take immediate action.

bbengfort: And the logic does make sense; this is saying that numOwners has to be 2 or greater (or negative I suppose … but that won't be returned from the query).

bbengfort (approved):
I mostly focused on the Quarterdeck database changes @pdeziel and @pdamodaran do you feel comfortable with the tests and the API handlers or would you like me to look at them more closely?

pkg/quarterdeck/api/v1/api.go
pdeziel: I think that's definitely better.

PR #413, opened 2023-04-17
Sc 16177/delete

daniellemaxwell:
### Scope of changes

Creates helper method to check if org has more than 1 owner before updating a member's role or deleting one. Also corrects a small grammar error in the teammate invitation email.

Fixes SC-16177 & SC-16163

### Type of change

- [ ] new feature
- [ ] bug fix
- [ ] documentation
- [ ] testing
- [ ] technical debt
- [ ] other (describe)

### Acceptance criteria

### Author checklist

### Reviewer(s) checklist


pkg/tenant/members_test.go
```diff
@@ -360,8 +360,6 @@ func (suite *tenantTestSuite) TestMemberDetail() {
 		switch in.Namespace {
 		case db.MembersNamespace:
 			return &pb.GetReply{Value: data}, nil
-		case db.OrganizationNamespace:
```
daniellemaxwell, line 363: I removed mentions of `OrganizationNamespace` since org verification has been removes from the member handlers.

pkg/tenant/members.go
```diff
@@ -467,3 +502,42 @@ func (s *Server) InvitePreview(c *gin.Context) {
...
+	// If the member to be updated is an owner, loop over dbMember and break out of the loop if there are at least two owners.
+	// If member is the only owner, their role cannot be changed.
+	if role == perms.RoleOwner {
+		// Get members from the database and set page size to return all members.
+		// TODO: Create list method that will not require pagination for this endpoint.
+		getAll := &pg.Cursor{StartIndex: "", EndIndex: "", PageSize: 100}
+		var members []*db.Member
+		if members, _, err = db.ListMembers(c, orgID, getAll); err != nil {
+			sentry.Error(c).Err(err).Msg("could not list members")
+			c.JSON(http.StatusInternalServerError, api.ErrorResponse("could not get member from the database"))
```
pdeziel, line 519: To simplify this further, I would prefer if we would not do any logging or JSON responses in this method (save it for the handler). Perhaps this method can just return the number of owners and we can have the `switch count` on the handler itself?

pdeziel (changes requested):
Good start, I had a few suggestions to simplify the checks and the error handling.

pkg/tenant/members.go
```diff
@@ -361,6 +361,17 @@ func (s *Server) MemberRoleUpdate(c *gin.Context) {
 		return
 	}
 
+	// Check to ensure the memberID from the URL matches the member ID from the database.
+	if memberID != member.ID {
+		c.JSON(http.StatusInternalServerError, api.ErrorResponse("member id does not match id in URL"))
```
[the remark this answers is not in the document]
daniellemaxwell: This link from a previous PR explains why this is currently a 500: https://github.com/rotationalio/ensign/pull/363/files/eab1cd9074222303a4b8518d1f019a98406fc0a3#r1156127767

I thought a 400 worked best, but the rational for using a 500 also made sense.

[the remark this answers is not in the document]
pdeziel: Ok I misread this the first time. Since we already used the memberID to retrieve the member record, I would vote for just removing this check.

pkg/tenant/members.go
```diff
@@ -361,6 +361,41 @@ func (s *Server) MemberRoleUpdate(c *gin.Context) {
...
+		getAll := &pg.Cursor{StartIndex: "", EndIndex: "", PageSize: 100}
+		var members []*db.Member
+		if members, _, err = db.ListMembers(c, orgID, getAll); err != nil {
+			sentry.Error(c).Err(err).Msg("could not list members")
+			c.JSON(http.StatusInternalServerError, api.ErrorResponse("could not get member from the database"))
+			return
+		}
+
+		count := uint8(0)
+		if count = orgOwnerCount(members); err != nil {
```
pdeziel, line 384: To be honest it feels weird to have to pass in the list of members into `orgOwnerCount`, can we have the helper do that work or was there an additional problem with that?

pkg/tenant/members.go
```diff
@@ -361,6 +361,41 @@ func (s *Server) MemberRoleUpdate(c *gin.Context) {
...
+		// TODO: Create list method that will not require pagination for this endpoint.
+		getAll := &pg.Cursor{StartIndex: "", EndIndex: "", PageSize: 100}
+		var members []*db.Member
+		if members, _, err = db.ListMembers(c, orgID, getAll); err != nil {
+			sentry.Error(c).Err(err).Msg("could not list members")
+			c.JSON(http.StatusInternalServerError, api.ErrorResponse("could not get member from the database"))
+			return
+		}
+
+		count := uint8(0)
```
pdeziel, line 383: Any particular reason why this is a `uint8`? Since it's a temporary variable I think it's okay to be a normal, signed int. If we were storing it, we might consider making it unsigned.

pdeziel (commented):
I've completed a second pass, let me know if you need to discuss anything.

pkg/tenant/members.go
daniellemaxwell: Are you suggesting that `ListMembers' return to the helper? I removed it after the comment about removing all JSON responses from the method.

pkg/tenant/members.go
daniellemaxwell: I was thinking of being more specific about the type of number that should be returned here, but this can be an `int`.

pdeziel: Alright, I don't have too strong of an opinion on it, but keep in mind that with `uint8` it will have a max value of 255, maybe just to be safe we should use `uint16`.

Also since the count is being returned from the method we could just do:

`var count uint8`

instead of manually initializing it to 0.

pkg/tenant/members.go
pdeziel: We can return a JSON error if `orgOwnerCount` returns an error.

pkg/tenant/members.go
daniellemaxwell: The loop should break at 2 owners, so that's why I went with uint8. It's not a big thing, so I'll use `int`.

pkg/tenant/members.go
```diff
@@ -361,6 +361,29 @@ func (s *Server) MemberRoleUpdate(c *gin.Context) {
...
+		if count, err = orgOwnerCount(c.Request.Context(), orgID); err != nil {
+			sentry.Error(c).Err(err).Msg("could not list members")
+			c.JSON(http.StatusInternalServerError, api.ErrorResponse("could not update member role"))
+			return
+		}
+
+		switch count {
+		case 0:
+			sentry.Warn(c).Err(err).Msg("could not find any owners")
+			c.JSON(http.StatusInternalServerError, api.ErrorResponse("could not get member from the database"))
```
pdeziel, line 379:
```suggestion
			c.JSON(http.StatusInternalServerError, api.ErrorResponse("could not update member role"))
```

pkg/tenant/members.go
```diff
@@ -420,6 +443,40 @@ func (s *Server) MemberDelete(c *gin.Context) {
...
+		if count, err = orgOwnerCount(c.Request.Context(), member.OrgID); err != nil {
+			sentry.Error(c).Err(err).Msg("could not list members")
+			c.JSON(http.StatusInternalServerError, api.ErrorResponse("could not delete member"))
+			return
+		}
+
+		switch count {
+		case 0:
+			sentry.Warn(c).Err(err).Msg("could not find any owners")
+			c.JSON(http.StatusInternalServerError, api.ErrorResponse("could not get member from the database"))
```
pdeziel, line 472:
```suggestion
			c.JSON(http.StatusInternalServerError, api.ErrorResponse("could not delete member"))
```

pdeziel (approved):
Looks good, thanks for bearing with the changes! Just a few error string suggestions, hopefully you can apply those directly in github.

PR #420, opened 2023-04-17
Switch Organization Authentication Endpoint

bbengfort:
### Scope of changes

Implements a `/v1/switch` endpoint so that authenticated users can switch organizations without having to login again.

Fixes SC-16223

### Type of change

- [x] new feature
- [ ] bug fix
- [ ] documentation
- [ ] testing
- [ ] technical debt
- [ ] other (describe)

### Acceptance criteria

Please see questions in the comments line by line.

### Author checklist

### Reviewer(s) checklist

- [ ] I've resolved all open questions created by @bbengfort


pkg/quarterdeck/auth.go
```diff
@@ -637,6 +639,126 @@ func (s *Server) Refresh(c *gin.Context) {
...
+	}
+
+	// If the orgID in the claims is the same as the requested orgID return an error
+	// (the user must use the refresh endpoint to get claims for the same orgID)
+	if orgID := claims.ParseOrgID(); orgID.Compare(in.OrgID) == 0 {
+		sentry.Warn(c).Msg("user attempting to switch to the same organization")
+		c.JSON(http.StatusBadRequest, api.ErrorResponse("cannot switch into the organization you are currently logged into"))
+		metrics.FailedLogins.WithLabelValues(ServiceName, UserHuman, "switch to same org").Inc()
+		return
+	}
```
bbengfort, line 690: I debated this - I would hate for the `switch` endpoint to be used for refreshing tokens; but this would require the front-end to filter the list of organizations so that the user couldn't click on the organization they're already logged into. Any thoughts on this check?

pkg/quarterdeck/auth.go
```diff
@@ -637,6 +639,126 @@ func (s *Server) Refresh(c *gin.Context) {
...
+			c.JSON(http.StatusForbidden, api.ErrorResponse("invalid credentials"))
+			metrics.FailedLogins.WithLabelValues(ServiceName, UserHuman, "invalid credentials").Inc()
+			return
+		}
+
+		sentry.Error(c).Err(err).Msg("could not retrieve organization user from database")
+		c.JSON(http.StatusInternalServerError, api.ErrorResponse("could not process switch request"))
+		metrics.FailedLogins.WithLabelValues(ServiceName, UserHuman, "unhandled error").Inc()
+		return
+	}
```
bbengfort, line 706: Two methods here:

1. retrieve org user from database (As done here)
2. retrieve user and call `SwitchOrganization`

I debated on which one to use; `SwitchOrganization` is used in Login so now I'm leaning towards that. Any thoughts?

pkg/quarterdeck/auth.go
```diff
@@ -637,6 +639,126 @@ func (s *Server) Refresh(c *gin.Context) {
...
+
+	// Add the user permissions to the claims
+	if claims.Permissions, err = user.Permissions(c.Request.Context(), false); err != nil {
+		sentry.Error(c).Err(err).Msg("could not get user permissions from model")
+		c.JSON(http.StatusInternalServerError, api.ErrorResponse("could not create credentials"))
+
+		// increment failure count
+		metrics.FailedLogins.WithLabelValues(ServiceName, UserHuman, "permissions not found").Inc()
+		return
+	}
```
bbengfort, line 736: It's _really_ important that new claims are created and that the old claims are NOT reused; should I stress that more in comments here; should I make it more secure so that we don't accidentally put in a bug with the old claims?

pkg/quarterdeck/auth.go
```diff
@@ -637,6 +639,126 @@ func (s *Server) Refresh(c *gin.Context) {
...
+		// increment failure count
+		metrics.FailedLogins.WithLabelValues(ServiceName, UserHuman, "jwt token error").Inc()
+		return
+	}
+
+	// Update the user's last login in a Go routine so it doesn't block
+	s.tasks.QueueContext(sentry.CloneContext(c), tasks.TaskFunc(func(ctx context.Context) error {
+		ctx, cancel := context.WithTimeout(ctx, 1*time.Minute)
+		defer cancel()
+		return user.UpdateLastLogin(ctx)
```
bbengfort, line 754: Do we have to update the organization last login as well?

pkg/quarterdeck/auth.go
```diff
@@ -637,6 +639,126 @@ func (s *Server) Refresh(c *gin.Context) {
...
+	}), tasks.WithError(fmt.Errorf("could not update last login timestamp for user %s", user.ID.String())))
+
+	// increment active users (in grafana we will divide by 24 hrs to get daily active)
+	metrics.Active.WithLabelValues(ServiceName, UserHuman).Inc()
+	c.JSON(http.StatusOK, out)
+}
```
bbengfort, line 760: TODO: need to write tests for the endpoint

pkg/quarterdeck/quarterdeck.go
```diff
@@ -253,6 +253,9 @@ func (s *Server) Routes(router *gin.Engine) (err error) {
 		v1.POST("/refresh", s.Refresh)
 		v1.POST("/verify", s.VerifyEmail)
 
+		// Authenticated access routes
+		v1.POST("/switch", authenticate, s.Switch)
```
bbengfort, line 257: Critical: this is an authenticated endpoint.

pkg/quarterdeck/auth.go
pdeziel: The frontend has access to the orgID in the claims right? So my hope is that filtering the list isn't too difficult, and I think it would be potentially confusing for the user to see their current org in the list if the option is called "switch". Preventing the switch to the same organization seems like the right behavior to me.

pkg/quarterdeck/auth.go
pdeziel: Perhaps we should use the `SwitchOrganization` since it abstracts a bit more? Then this method doesn't have to know what an `OrganizationUser` is.

pkg/quarterdeck/auth.go
pdeziel: Am I missing something or is there already a bug? Don't the permissions need to be copied into `newClaims`? Maybe one way to make this more foolproof is to create a method on the `User` which constructs the new claims.

pkg/quarterdeck/auth.go
pdeziel: This method was modified to also update the last login on the user's "loaded" organization, so we should be covered.

pdeziel (commented):
Looks good so far, I had a few suggestions relating to your comments so let me know what you think.

pkg/quarterdeck/auth.go
bbengfort: Awesome; then we'll leave this in. Would you mind helping me make sure that Edwin and Cletus understand this functionality?

pkg/quarterdeck/auth.go
bbengfort: Ok sounds good; I'll change the code to use `SwitchOrganization`. Will try to do it when I make it to Miami.

pkg/quarterdeck/auth.go
bbengfort: The bug would be if anything was copied from the old claims to the new claims. The permissions must come from the new organization role; e.g. if the user is an Owner in Organization A but a member in Organization B; the new claims must reflect the role in B if they're switching from A to B; replacing the claims from A.

pkg/quarterdeck/auth.go
bbengfort: Perfect!

bbengfort:
@pdeziel ok, wrapped up this PR -- would you mind merging if the review is ok?

pdeziel (approved):

PR #421, opened 2023-04-18
Adds organization list to tenant

daniellemaxwell:
### Scope of changes

Adds `OrganizationList` to Tenant.

Fixes SC-16170

### Type of change

- [x] new feature
- [ ] bug fix
- [ ] documentation
- [ ] testing
- [ ] technical debt
- [ ] other (describe)

### Acceptance criteria

### Author checklist

### Reviewer(s) checklist

### DIFF
--- a/pkg/tenant/api/v1/api.go
+++ b/pkg/tenant/api/v1/api.go
@@ -18,6 +18,7 @@ type TenantClient interface {
 	VerifyEmail(context.Context, *VerifyRequest) error
 	InvitePreview(context.Context, string) (*MemberInvitePreview, error)
 
+	OrganizationList(context.Context, *PageQuery) (*OrganizationPage, error)
 	OrganizationDetail(context.Context, string) (*Organization, error)
 
 	TenantList(context.Context, *PageQuery) (*TenantPage, error)
@@ -179,13 +180,19 @@ type PageQuery struct {
 }
 
 type Organization struct {
-	ID       string `json:"id" uri:"id"`
-	Name     string `json:"name"`
-	Owner    string `json:"owner"`
-	Domain   string `json:"domain"`
-	Projects int    `json:"projects"`
-	Created  string `json:"created"`
-	Modified string `json:"modified"`
+	ID        string `json:"id" uri:"id"`
+	Name      string `json:"name"`
+	Owner     string `json:"owner"`
+	Domain    string `json:"domain"`
+	Projects  int    `json:"projects"`
+	LastLogin string `json:"last_login"`
+	Created   string `json:"created"`
+	Modified  string `json:"modified"`
+}
+
+type OrganizationPage struct {
+	Organizations []*Organization `json:"organizations"`
+	NextPageToken string          `json:"next_page_token,omitempty"`
 }
 
 type Tenant struct {
--- a/pkg/tenant/organizations.go
+++ b/pkg/tenant/organizations.go
@@ -19,7 +19,7 @@ import (
 	"github.com/rotationalio/ensign/pkg/utils/ulids"
 )
 
-// OrganizationList fetches a list of organizations from Quarterdeck.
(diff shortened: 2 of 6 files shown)


pkg/tenant/api/v1/api.go
```diff
@@ -180,13 +181,19 @@ type PageQuery struct {
...
-	ID       string `json:"id" uri:"id"`
-	Name     string `json:"name"`
-	Owner    string `json:"owner"`
-	Domain   string `json:"domain"`
-	Projects int    `json:"projects"`
-	Created  string `json:"created"`
-	Modified string `json:"modified"`
+	ID        string `json:"id" uri:"id"`
+	Name      string `json:"name"`
+	Owner     string `json:"owner"`
```
pdeziel, line 186: I think we should get rid of the `Owner` field, I know it was already there but an organization can have multiple members with the `Owner` role, so it doesn't really make sense anymore.

pkg/tenant/organizations.go
```diff
@@ -19,6 +19,64 @@ import (
 	"github.com/rotationalio/ensign/pkg/utils/ulids"
 )
 
+// OrganizationList fetches a list of organizations from Quarterdeck.
```
pdeziel, line 22:
```suggestion
// OrganizationList fetches the list of organizations the authenticated user is a part of from Quarterdeck.
```

pkg/tenant/organizations.go
```diff
@@ -19,6 +19,64 @@ import (
...
+	}
+
+	for _, org := range reply.Organizations {
+		orgs := &api.Organization{
+			ID:        org.ID.String(),
+			Name:      org.Name,
+			Domain:    org.Domain,
+			Created:   db.TimeToString(org.Created),
+			LastLogin: db.TimeToString(org.LastLogin),
+		}
```
pdeziel, line 73: Now that we are converting the DB model into an API model in two places, do you mind creating a `ToAPI` method in `pkg/tenant/db/organizations.go` similar to the one for the other resources? It wasn't necessary before but is now more important since we want to make sure the response is the same for both `OrganizationList` and `OrganizationDetail`.

While we're at it, we should also use this method in `OrganizationDetail` to make sure that we're populating `LastLogin` there as well.

pdeziel (approved):
Very nice pass through endpoint. I didn't have much to add besides a small organizational suggestion. Endpoint and tests look great and complete!

pkg/tenant/api/v1/api.go
daniellemaxwell: The Figma designs display an owner, and that's currently what appears on the organization page so should we leave this put for now? The `getOwner` helper method used in `OrganizationDetails` returns the first member with that role for now.

pkg/tenant/organizations.go
daniellemaxwell: The only db model we have for `Organizations` is in Quarterdeck. We're not storing org details in trtl, so it feels like we adding one isn't necessary. Also, the Figma designs only display `Created` in `OrganizationDetail` and the `Organizations` table, so `LastLogin` would not be displayed anywhere, at least for now, if it's returned. If I recall, we're using `LastLogin` here to make it easier to return a user to the last org they were logged into.

pkg/tenant/api/v1/api.go
pdeziel: Oh, I completely forgot about that. Can we create a spike story to reassess what the owner field should be?

pdeziel: We can leave the field in for right now.

pkg/tenant/organizations.go
pdeziel: I see, we may want to have a helper method eventually but for the scope of this PR I think it's okay to go with what you have.

PR #430, opened 2023-04-21
Tenant return confirm on delete

pdeziel:
### Scope of changes

This updates the Tenant member delete so that it makes a pass through request to Quarterdeck to attempt to remove the user from the organization, and returns the confirmation token if required.

Fixes SC-16121

### Type of change

- [x] new feature
- [ ] bug fix
- [ ] documentation
- [ ] testing
- [ ] technical debt
- [ ] other (describe)

### Acceptance criteria

### Author checklist

### Reviewer(s) checklist


pkg/tenant/members.go
pdeziel, line 1: One potential problem I'm realizing is that with the "delete" and "delete confirm" endpoints the user will have to actually confirm twice in the confirm case. Once to initially try to delete the user and another time after reviewing the list of resources that would be deleted. Are we okay with this?

bbengfort: I am ok with this - though we can certainly change the messaging on the UI to make it a bit more streamlined. Let's bring it up in the UI sync this week with Edwin to make sure he's on the same page as us.

bbengfort: Also; the first confirmation is up to the UI so we could remove it if necessary (e.g. there is no token if the user doesn't have any associated resources).

bbengfort (approved):

PR #436, opened 2023-04-24
Humanize Quarterdeck and Tenant Error Messages

rebeccabilbro:
### Scope of changes

This PR both standardizes and humanizes the errors that have a high potential of getting passed from either Quarterdeck or Tenant up through to a customer via Gin. We're treading a fine line between giving people enough information to know how to proceed with their Beacon action and exposing info that could create security holes, so I geared the error messages away from providing reasons for the error and more towards providing coaching to a solution.

Fixes SC-15386

### Type of change

- [ ] new feature
- [ ] bug fix
- [ ] documentation
- [ ] testing
- [ ] technical debt
- [x] other (nicer error messages for better Beacon error UX)

### Acceptance criteria

### Author checklist

### Reviewer(s) checklist


pkg/quarterdeck/errors.go
```diff
@@ -0,0 +1,14 @@
+package quarterdeck
+
+// String errors intended to pass back from the server up to
+// a **human** Beacon user so these need to make sense to our customers!
+var (
+	ErrTryLoginAgain      = "Unable to login with those details - please try again!"
+	ErrTryProjectAgain    = "Unable to create or access that project - please try again!"
+	ErrNeedPermission     = "Unable to create or access projects - please request permission from your team owner."
```
daniellemaxwell, line 8: Should this message use `organization` instead of `team`?

pkg/quarterdeck/orgs.go
```diff
@@ -159,21 +160,21 @@ func (s *Server) ProjectCreate(c *gin.Context) {
...
-		c.JSON(http.StatusBadRequest, api.ErrorResponse(api.ErrUnparsable))
+		c.JSON(http.StatusBadRequest, api.ErrorResponse(responses.ErrTryProjectAgain))
 		return
 	}
 
 	// Validate the request from the API side.
 	if err = project.Validate(); err != nil {
 		c.Error(err)
-		c.JSON(http.StatusBadRequest, api.ErrorResponse(err))
+		c.JSON(http.StatusBadRequest, api.ErrorResponse(responses.ErrTryProjectAgain))
```
daniellemaxwell, line 170: Since this is an error related to validation, would it be helpful to have a message informing users a project can't be created with the details provided, in a way similar to `ErrTryLoginAgain`, or would that give too much away?

pkg/quarterdeck/orgs.go
```diff
@@ -159,21 +160,21 @@ func (s *Server) ProjectCreate(c *gin.Context) {
...
 		return
 	}
 
 	// Fetch the user claims from the request
 	if claims, err = middleware.GetClaims(c); err != nil {
 		sentry.Error(c).Err(err).Msg("could not get user claims from authenticated request")
-		c.JSON(http.StatusUnauthorized, api.ErrorResponse(api.ErrInvalidUserClaims))
+		c.JSON(http.StatusUnauthorized, api.ErrorResponse(responses.ErrNeedPermission))
```
daniellemaxwell, line 177: Would `ErrSomethingWentWrong` be a little more appropriate here since an error is returned if the user's claims aren't fetched?

pkg/quarterdeck/orgs.go
```diff
@@ -233,20 +234,20 @@ func (s *Server) ProjectAccess(c *gin.Context) {
...
```
daniellemaxwell, line 237: Same here as the comment above.

web/beacon-app/src/features/members/types/addMemberFormService.ts
```diff
@@ -12,7 +12,7 @@ export const FORM_INITIAL_VALUES = {
 } satisfies NewMemberDTO;
 
 export const FORM_VALIDATION_SCHEMA = object({
-  email: string().trim().required('Email is required').email('Email is invalid'),
+  email: string().trim().required('Email is required').email('Email is invalid.'),
```
daniellemaxwell, line 15:
```suggestion
  email: string().trim().required('Email is required.').email('Email is invalid.'),
```

web/beacon-app/src/features/members/types/addMemberFormService.ts
```diff
@@ -12,7 +12,7 @@ export const FORM_INITIAL_VALUES = {
...
   role: string().required('role is required'),
```
daniellemaxwell, line 16:
```suggestion
  role: string().required('Role is required.'),
```

pkg/quarterdeck/auth.go
```diff
@@ -589,12 +590,13 @@ func (s *Server) Refresh(c *gin.Context) {
...
 		if errors.Is(err, models.ErrUserOrganization) {
-			c.JSON(http.StatusForbidden, api.ErrorResponse("invalid credentials"))
+			c.JSON(http.StatusForbidden, api.ErrorResponse(responses.ErrLogBackIn))
 			return
 		}
 
 		sentry.Warn(c).Err(err).Msg("could not retrieve user from database")
-		c.JSON(http.StatusForbidden, api.ErrorResponse("could not retrieve user from claims"))
+		// Should this be an http.StatusInternalServerError instead? :point-down:
+		c.JSON(http.StatusForbidden, api.ErrorResponse(responses.ErrLogBackIn))
```
daniellemaxwell, line 599: I'm assuming `http.StatusForbidden` is currently used here because if a user can't be retrieved from the database, then they shouldn't have access. If the user is told to log in again, then I think that `http.StatusInternalServerError` works better because we would not want to inform a user that they should try logging in again while also returning a forbidden error.

pkg/quarterdeck/orgs.go
rebeccabilbro: Hmm, possibly; it might depend on what triggers the validation failures in practice, probably Beacon should be doing form validation before sending requests to QD or Tenant, and those validation errors can help guide the user to fill in fields that they've left empty.

pkg/quarterdeck/orgs.go
rebeccabilbro: Does that mean the error should be `http.StatusInternalServerError` instead of `http.StatusUnauthorized`?

pkg/quarterdeck/orgs.go
rebeccabilbro: Same question as above - if the failure is due to something internal that Ensign has done wrong, should the error be `StatusInternalServerError` instead of `StatusUnauthorized`?

daniellemaxwell (commented):
Thank you for providing more actionable error messages to users as I'm sure they will appreciate it! I have a couple of questions that may require a little more discussion. Although, I just saw your reply to a couple, so I'll submit this and then follow up.

pkg/quarterdeck/orgs.go
daniellemaxwell: I took a look at the `GetClaims` method and it returns an error if no claims exist on the context, so I feel that's why `http.StatusUnauthorized` is used instead of `http.StatusInternalServerError`. In that case, informing the user they need permission seems fine, but should they be directed to their owner in this scenario?

daniellemaxwell (approved):
I replied to the last comment about the error message related to parsing claims and made some minor suggestions to the FE errors. Thanks for going through all of this and also making comments where additional changes could be made.

pkg/quarterdeck/errors.go
rebeccabilbro: @daniellemaxwell we should use whichever term is used in the UI (i.e. if the UI calls them "Teams", this error message should reference "Teams"). So I think the real problem is that the UI references them as "Teams" in some places and "Organizations" in others. I've [created a story to address this](https://app.shortcut.com/rotational/story/16474/choose-either-teams-or-organizations-and-update-beacon-copy).

PR #442, opened 2023-04-25
Add projects stat fields

pdeziel:
### Scope of changes

This adds some stat fields to the project model in Tenant, and adds a method to update them asynchronously. Eventually we want to have the updates populated via Ensign so the method is more of a stop gap measure for now.

Fixes SC-16445

### Type of change

- [x] new feature
- [ ] bug fix
- [ ] documentation
- [ ] testing
- [ ] technical debt
- [ ] other (describe)

### Acceptance criteria

### Author checklist

### Reviewer(s) checklist


pkg/tenant/apikeys.go
```diff
@@ -225,6 +227,11 @@ func (s *Server) ProjectAPIKeyCreate(c *gin.Context) {
 		Modified:     key.Modified.Format(time.RFC3339Nano),
 	}
 
+	// Update project stats in the background
+	s.tasks.QueueContext(sentry.CloneContext(c), tasks.TaskFunc(func(ctx context.Context) error {
+		return s.UpdateProjectStats(ctx, key.ProjectID)
+	}), tasks.WithError(fmt.Errorf("could not update stats for project %s", key.ProjectID.String())))
```
pdeziel, line 233: Only doing this in `ProjectAPIKeyCreate` and `ProjectTopicCreate` for right now so the status shows up properly in the frontend. I'm hoping that we can replace with the ensign topic once Tenant is setup for that.

pkg/tenant/projects.go
```diff
@@ -514,3 +514,67 @@ func (s *Server) createProject(ctx context.Context, project *db.Project) (err er
...
+			return err
+		}
+
+		numAPIKeys += uint64(len(reply.APIKeys))
+		if reply.NextPageToken == "" {
+			break
+		}
+
+		req.NextPageToken = reply.NextPageToken
+	}
```
bbengfort, line 547: Yeh, I can see the issue; this isn't particularly performant -- probably what we should do is have a Quarterdeck endpoint that returns the number of API keys -- I'll create a story for that.

pkg/tenant/projects.go
```diff
@@ -514,3 +514,67 @@ func (s *Server) createProject(ctx context.Context, project *db.Project) (err er
...
+
+	// Count the number of topics in trtl.
+	var next *pg.Cursor
+	for {
+		var topics []*db.Topic
+		if topics, next, err = db.ListTopics(ctx, projectID, next); err != nil {
+			return err
+		}
+
+		numTopics += uint64(len(topics))
```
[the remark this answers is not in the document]
bbengfort: We probably would want to differentiate them in the tenant stats. We probably also want an Ensign endpoint for this as well.

pkg/tenant/projects.go
```diff
@@ -514,3 +514,67 @@ func (s *Server) createProject(ctx context.Context, project *db.Project) (err er
...
+	}
+
+	// Retrieve the project from trtl.
+	var project *db.Project
+	if project, err = db.RetrieveProject(ctx, projectID); err != nil {
+		return err
+	}
+
+	// TODO: A project write in between here could cause updates to be stomped (e.g. if
+	// the project name is updated it will be overwritten here).
```
bbengfort, line 570: This is a good point and there needs to be a transaction read then write in trtl to allow for a read then a write.

bbengfort (commented):
This does make sense to me - though I think we need to streamline and parallelize the Quarterdeck and Ensign queries. Let me see if I can quickly create endpoints for project info -- then in the meantime you can perform the data fetching in a go routine?

cmd/debug/main.go
```diff
@@ -208,7 +208,7 @@ func generate(c *cli.Context) (err error) {
 				}
 
 				msg++
-				publisher.Publish(&api.Event{
+				publisher.Publish("generator", &api.Event{
```
pdeziel, line 211: Can you verify this is the correct usage? I didn't run the script to test it.

pkg/tenant/projects.go
```diff
@@ -514,3 +517,96 @@ func (s *Server) createProject(ctx context.Context, project *db.Project) (err er
...
+	rpc := func(fn func(context.Context) (uint64, error), i int, res *uint64) {
+		defer wg.Done()
+		*res, errs[i] = fn(ctx)
+	}
+	wg.Add(2)
+	go rpc(countAPIKeys, 0, numAPIKeys)
+	go rpc(countTopics, 1, numTopics)
+	wg.Wait()
+
+	// Return if both RPCs errored.
```
pdeziel, line 583: Should we return if only one of them errors to keep the Tenant view more consistent?

cmd/debug/main.go
bbengfort: I think that needs to be a topic id - but don't worry about it; we'll need to add authentication etc. Adding "generator" here ensures that ensign builds which is enough for now.

pkg/tenant/projects.go
```diff
@@ -514,3 +517,96 @@ func (s *Server) createProject(ctx context.Context, project *db.Project) (err er
...
+	errs := make([]error, 2)
+	res := make([]uint64, 2)
+	rpc := func(fn func(context.Context) (uint64, error), i int) {
+		defer wg.Done()
+		res[i], errs[i] = fn(ctx)
+	}
+	wg.Add(2)
+	go rpc(countAPIKeys, 0)
+	go rpc(countTopics, 1)
+	wg.Wait()
```
bbengfort, line 581: Nicely concise!

pkg/tenant/projects.go
bbengfort: Yes, you're probably right; e.g. if the user doesn't have read:topics we still want to know how many api keys.

bbengfort (approved):
Looks good - glad this is a bit easier than listing everything from Ensign and Quarterdeck!

PR #452, opened 2023-04-27
Adds description to project api struct and model

daniellemaxwell:
### Scope of changes

Adds `description` to `Project` API struct and model. Beacon users will have an option to include a description when creating a project. If a description is provided, it will be included in the `Project Detail` tooltip.

Links to the Figma designs are below for reference:
https://www.figma.com/file/PvL23fm6W36Zd88O1XVyIo/Beacon-DX-UI?node-id=4631-25864&t=T9NuNKp8hwnJhaNi-4

https://www.figma.com/file/PvL23fm6W36Zd88O1XVyIo/Beacon-DX-UI?node-id=5374-14608&t=T9NuNKp8hwnJhaNi-4

Fixes SC-16585

### Type of change

- [x] new feature
- [ ] bug fix
- [ ] documentation
- [ ] testing
- [ ] technical debt
- [ ] other (describe)

### Acceptance criteria

### Author checklist

### Reviewer(s) checklist

### DIFF
--- a/pkg/tenant/db/projects.go
+++ b/pkg/tenant/db/projects.go
@@ -16,12 +16,13 @@ import (
 const ProjectNamespace = "projects"
 
 type Project struct {
-	OrgID    ulid.ULID `msgpack:"org_id"`
-	TenantID ulid.ULID `msgpack:"tenant_id"`
-	ID       ulid.ULID `msgpack:"id"`
-	Name     string    `msgpack:"name"`
-	Created  time.Time `msgpack:"created"`
-	Modified time.Time `msgpack:"modified"`
+	OrgID       ulid.ULID `msgpack:"org_id"`
+	TenantID    ulid.ULID `msgpack:"tenant_id"`
+	ID          ulid.ULID `msgpack:"id"`
+	Name        string    `msgpack:"name"`
+	Description string    `msgpack:"description"`
+	Created     time.Time `msgpack:"created"`
+	Modified    time.Time `msgpack:"modified"`
 }
 
 var _ Model = &Project{}
@@ -72,10 +73,11 @@ func (p *Project) Validate() error {
 // Convert the model to an API response.
 func (p *Project) ToAPI() *api.Project {
 	return &api.Project{
-		ID:       p.ID.String(),
-		Name:     p.Name,
-		Created:  TimeToString(p.Created),
-		Modified: TimeToString(p.Modified),
+		ID:          p.ID.String(),
+		Name:        p.Name,
+		Description: p.Description,
+		Created:     TimeToString(p.Created),
+		Modified:    TimeToString(p.Modified),
 	}
 }
 
--- a/pkg/tenant/db/projects.go
+++ b/pkg/tenant/db/projects.go
@@ -67,6 +67,10 @@ func (p *Project) Validate() error {
 		return ErrMissingProjectName
 	}
 
+	if len(p.Description) > 2000 {
+		return ErrProjectDescriptionTooLong
+	}
+
 	return nil
 }
 
--- a/pkg/tenant/db/projects.go
+++ b/pkg/tenant/db/projects.go
@@ -13,7 +13,10 @@ import (
 	"golang.org/x/net/context"
 )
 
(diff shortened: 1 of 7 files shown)


pkg/tenant/projects.go
```diff
@@ -156,6 +156,10 @@ func (s *Server) TenantProjectCreate(c *gin.Context) {
 		Name:     project.Name,
 	}
 
+	if project.Description != "" {
```
daniellemaxwell, line 159: The Figma design doesn't include a max word count, but I'm wondering if we should add a check for that.

daniellemaxwell: Never mind. In SC-15723, the max length for the description is listed as 2k characters.

pdeziel (commented):
Looks great, I agree that the description should be optional. I think we should add the length check as you suggested to prevent the user from being able to add arbitrarily long descriptions, so I didn't mark this as approved yet.

pkg/tenant/db/projects.go
```diff
@@ -72,10 +73,11 @@ func (p *Project) Validate() error {
 // Convert the model to an API response.
 func (p *Project) ToAPI() *api.Project {
 	return &api.Project{
-		ID:       p.ID.String(),
-		Name:     p.Name,
-		Created:  TimeToString(p.Created),
-		Modified: TimeToString(p.Modified),
+		ID:          p.ID.String(),
+		Name:        p.Name,
+		Description: p.Description,
```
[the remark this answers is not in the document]
pdeziel: Can we also add an assertion in the project detail test to make sure that the description is being returned?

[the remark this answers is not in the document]
daniellemaxwell: Yes, that makes sense as the description will not need to be returned on list.

pkg/tenant/projects.go
```diff
@@ -424,6 +433,15 @@ func (s *Server) ProjectUpdate(c *gin.Context) {
 	// Update all user provided fields
 	p.Name = project.Name
 
+	if project.Description != "" {
```
daniellemaxwell, line 436: I added this after realizing that user's may want to edit their project description.

pkg/tenant/db/projects_test.go
```diff
@@ -67,8 +69,13 @@ func TestProjectValidate(t *testing.T) {
 	project.Name = " "
 	require.ErrorIs(t, project.Validate(), db.ErrMissingProjectName, "expected missing name error")
 
-	// Test valid project
+	// Test description that's too long
 	project.Name = "Hello World"
+	project.Description = strings.Repeat("a", 2001)
```
pdeziel, line 74: I did not know about `strings.Repeat`, nice test hack!

pkg/tenant/db/projects.go
```diff
@@ -66,16 +67,21 @@ func (p *Project) Validate() error {
 		return ErrMissingProjectName
 	}
 
+	if len(p.Description) > 2000 {
+		return ErrProjectDescriptionTooLong
```
pdeziel, line 71: Can you make sure that this constraint is mentioned in the frontend story for project create?

pkg/tenant/projects.go
```diff
@@ -156,6 +156,15 @@ func (s *Server) TenantProjectCreate(c *gin.Context) {
 		Name:     project.Name,
 	}
 
+	if project.Description != "" {
+		if len(project.Description) > 2000 {
```
pdeziel, line 160: We'll probably want to make this a constant so we can share it between api and db validation.

pkg/tenant/projects.go
pdeziel: Agreed, good call!

pdeziel (approved):
Looks good, I had one suggestion about changing the max length to a constant so we don't have to update it in three places, but other than that looks good!

pkg/tenant/db/projects.go
daniellemaxwell: Edwin included the constraint when he created the FE story.

PR #458, opened 2023-05-01
Add project owner

pdeziel:
### Scope of changes

This adds project ownership fields to Quarterdeck and Tenant. Every project must have an owner, in Quarterdeck this means that an `owner_id` must be associated with the project. In Tenant this means that the owner ID, name, and email must be on the project model. This will allow the frontend to know which projects are owned by which users and will allow Quarterdeck to keep track of project ownership in the future when users want to change the owner of a project or delete a user who owns a project.

Fixes SC-15721

### Type of change

- [x] new feature
- [ ] bug fix
- [ ] documentation
- [ ] testing
- [ ] technical debt
- [ ] other (describe)

### Acceptance criteria

See reviewer checklist

### Author checklist

### Reviewer(s) checklist

- [ ] Should the project owner be stored in Quarterdeck?


pkg/quarterdeck/db/migrations/0006_projects.sql
```diff
@@ -0,0 +1,15 @@
...
+
+-- Owner of the project.
+ALTER TABLE organization_projects ADD COLUMN owner_id BLOB NOT NULL DEFAULT '';
+
+UPDATE organization_projects
+SET owner_id = (
+    SELECT user_id FROM organization_users
+    WHERE organization_id = organization_projects.organization_id AND role_id = 1
+    ORDER BY created LIMIT 1)
+WHERE owner_id = '';
```
pdeziel, line 13: Do we need a test for this migration or do you think deploying to staging is good enough?

pkg/quarterdeck/db/models/projects.go
pdeziel, line 1: In the story we didn't mention if the owner should be a Quarterdeck thing or just a Tenant thing, but I thought it made sense to at least have the ID stored in Quarterdeck since it needs to lookup the user owned resources on the user delete.

pkg/quarterdeck/db/migrations/0006_projects.sql
bbengfort: Deploying to staging is probably good enough.

pkg/quarterdeck/db/models/projects.go
bbengfort: It does indeed. So our choices are:

1. Quarterdeck doesn't know about project owners and it's up to Tenant to not delete project owners
2. Make project owners editable in Quarterdeck

I'm leaning towards option 1 - especially if project owner doesn't have special permissions or claims. What do you think?

pkg/tenant/api/v1/api.go
```diff
@@ -250,13 +250,15 @@ type TenantProjectPage struct {
...
-	Description string `json:"description,omitempty"`
-	Status      string `json:"status"`
-	Created     string `json:"created,omitempty"`
-	Modified    string `json:"modified,omitempty"`
+	ID           string `json:"id" uri:"id"`
+	TenantID     string `json:"tenant_id"`
+	Name         string `json:"name"`
+	Description  string `json:"description,omitempty"`
+	OwnerName    string `json:"owner_name"`
+	OwnerPicture string `json:"owner_picture"`
```
bbengfort, line 258: Can we return a `Member` object instead that is only partially updated with name and picture? 

```suggestion
	Owner    Member `json:"owner"`
```

Feels like this would be more maintainable.

pkg/tenant/db/projects.go
```diff
@@ -25,7 +27,15 @@ const (
 	ProjectStatusArchived   = "Archived"
 )
 
+type Owner struct {
+	ID       ulid.ULID `msgpack:"owner_id"`
+	Name     string    `msgpack:"owner_name"`
+	Email    string    `msgpack:"owner_email"`
+	Gravatar string    `msgpack:"owner_gravatar"`
```
bbengfort, line 34: Is the `owner_` prefix necessary for `msgpack`? 

Is the intent to store this inline with the nosql document rather than as a reference to the `members` namespace in order to optimize project list fetches? If so, is there a plan with keeping Owner name/email/gravatar up to date with the Member ID? E.g. we can guarantee that the owner is a member and if member changes, owner will change?

Also Gravatar can be computed from email and doesn't need to be stored in the database.

pkg/quarterdeck/db/models/projects.go
pdeziel: I think option 1 makes more sense, I'm not sure what a project owner really means to Quarterdeck if they have the same permissions as another organization member with the owner role. It feels more in the Tenant domain to me, and I think it's fine if Tenant has to do the project owner check, and that way we don't have to worry about making sure Tenant and Quarterdeck agree on who the owner is.

pkg/tenant/api/v1/api.go
pdeziel: Yeah that makes sense, the `Member` object doesn't currently have a picture field but it probably makes sense to add it, maybe we want to return it on Member List as well?

pkg/tenant/db/members.go
```diff
@@ -103,12 +105,21 @@ func (m *Member) Validate() error {
 	return nil
 }
 
+func (m *Member) Picture() string {
+	if m.gravatar == "" {
+		m.gravatar = gravatar.New(m.Email, nil)
+	}
+
+	return m.gravatar
+}
```
bbengfort, line 114: Cheers!

pkg/tenant/db/projects.go
```diff
@@ -29,13 +30,15 @@ type Project struct {
...
 	ID          ulid.ULID `msgpack:"id"`
+	OwnerID     ulid.ULID `msgpack:"owner_id"`
 	Name        string    `msgpack:"name"`
 	Description string    `msgpack:"description"`
 	Archived    bool      `msgpack:"archived"`
 	APIKeys     uint64    `msgpack:"api_keys"`
 	Topics      uint64    `msgpack:"topics"`
 	Created     time.Time `msgpack:"created"`
 	Modified    time.Time `msgpack:"modified"`
+	owner       *Member
```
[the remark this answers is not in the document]
bbengfort: Yes, we should probably do that. Man, we need to do some work on trtl …

bbengfort (approved):
Makes sense to me!

PR #467, opened 2023-05-03
Multi-Message Publish and Subscribe Streams

bbengfort:
### Scope of changes

This PR updates the protocol buffers for Ensign to allow for multi-message publish and subscribe streams allowing a streaming context to include control flow messages such as initial authentication of a publisher and redirects or topic queries inside of the stream. 

**WARNING**: This PR will break all of the SDKs and is not backward compatible with Ensign v0.6.0 -- but it is important to finalizing the production release of Ensign. 

Fixes SC-16136

### Type of change

- [x] new feature
- [ ] bug fix
- [ ] documentation
- [ ] testing
- [x] technical debt
- [ ] other (describe)

### Acceptance criteria

### Author checklist

- [x]  I have recompiled and included new protocol buffers to reflect changes I made

### Reviewer(s) checklist


proto/api/v1beta1/ensign.proto
```diff
@@ -37,33 +37,57 @@ service Ensign {
...
+    }
+}
+
+// PublisherReply messages are sent back to publishers from the server. Generally they
+// are responses to receiving events (e.g. ack and nack) but the last message contains
 // information about the performance of the publisher and the topic itself.
-message Publication {
+// This message can be extended to allow more general interactions between the publisher
+// and the server including redirects, topic queries, etc.
+message PublisherReply {
```
pdeziel, line 64: So here we could potentially send back the topic mappings to help the SDKs?

proto/api/v1beta1/ensign.proto
```diff
@@ -85,28 +109,76 @@ message Ack {
...
 
-// OpenStream is the first message that should be sent in a Subscribe stream as it
-// defines what topic the client is subscribing to and how it will process the events.
+// OpenStream is the first message that should be sent in a Publish stream in order to
+// identify and authenticate the publisher. Optionally, the publisher can specify the
+// topics that they want to publish to in order to check that they are being sent to the
+// correct node.
 message OpenStream {
-    string consumer_id = 1;
+    string client_id = 1;
```
pdeziel, line 145: So the `client_id` is just generated by the client and has nothing to do with the API key?

proto/api/v1beta1/ensign.proto
```diff
@@ -85,28 +109,76 @@ message Ack {
...
 
-// CloseStream returns some basic stats and topic information to the publisher when the
-// stream is closed and provides feedback that the stream was closed successfully.
+// CloseStream returns some basic stats and topic information to the publisher or
+// subscriber when the stream is closed and provides feedback that the stream was closed
+// successfully.
 message CloseStream {
     uint64 events = 1;
-    uint64 topic_offset = 2;
+    uint64 topics = 2;
```
pdeziel, line 154: What count does `topics` refer to here?

proto/api/v1beta1/event.proto
```diff
@@ -2,38 +2,121 @@ syntax = "proto3";
...
 import "mimetype/v1beta1/mimetype.proto";
 import "google/protobuf/timestamp.proto";
 
+// EventWrapper is a header wrapper for processing events by the Ensign server. It is
+// intended to reduce the processing time of events to increase throughput on the server
+// even though it will require more work from the SDKs to serialize events. However,
+// this will happen on the client-side making Ensign nodes much more parallel. The only
+// data in the EventWrapper should be data that is required for the Ensign node to
+// process events.
+message EventWrapper {
```
pdeziel, line 15: So for right now, the SDKs mostly just have to set the `topic_id` and serialize the `Event` into the `event` field?

pdeziel (commented):
I had a few questions about how specific fields are meant to be used, looks good!

proto/api/v1beta1/ensign.proto
bbengfort: Yes - any general interaction in the pub/sub stream could be wrapped in these messages.

proto/api/v1beta1/ensign.proto
bbengfort: That's correct - because multiple clients could be using the same API key.

proto/api/v1beta1/ensign.proto
bbengfort: Good question; I think it's the number of topics were actually published to or consumed from in the course of the stream? These metrics are a bit fluid right now since the backend is being updated so frequently. Any requests?

proto/api/v1beta1/event.proto
bbengfort: That's correct -- and possibly the `key` for sharding topics. I was definitely thinking server-side for most of these fields.

bbengfort: Also note that I made the topic_id the ULID bytes instead of ULID string to make the event wrapper more compact.

PR #471, opened 2023-05-03
Tenant ensign authentication

pdeziel:
### Scope of changes

This adds an Ensign client to Tenant that uses the Go SDK to make authenticated requests from handlers.

Fixes SC-15034

### Type of change

- [x] new feature
- [ ] bug fix
- [ ] documentation
- [ ] testing
- [ ] technical debt
- [ ] other (describe)

### Acceptance criteria

### Author checklist

### Reviewer(s) checklist


pkg/tenant/topics.go
```diff
@@ -443,27 +439,20 @@ func (s *Server) TopicUpdate(c *gin.Context) {
...
-		var tombstone *pb.TopicTombstone
-		if tombstone, err = s.ensign.DeleteTopic(ensignContext, updateRequest); err != nil {
-			if status.Code(err) == codes.NotFound {
-				sentry.Warn(c).Err(err).Str("topicID", updateRequest.Id).Msg("topic not found in ensign even though it is in tenant")
-				c.JSON(http.StatusNotFound, api.ErrorResponse("topic not found"))
+		if err = s.ensign.InvokeOnce(rep.AccessToken).ArchiveTopic(ctx, t.ID.String()); err != nil {
+			switch {
+			case err.Error() == "not implemented yet":
+				sentry.Warn(c).Err(err).Msg("this version of the Go SDK does not support topic archiving")
+				c.JSON(http.StatusNotImplemented, api.ErrorResponse("archiving a topic is not supported"))
```
pdeziel, line 447: The Go SDK doesn't currently support topic delete or topic archive so this is a temporary workaround for the tests. It would also be nice to have standardized errors from the SDK, not sure if there's a story for that yet.

pkg/tenant/client.go
```diff
@@ -0,0 +1,46 @@
...
+	}
+
+	return &EnsignClient{client: client, opts: opts}, nil
+}
+
+// InvokeOnce exposes a clone of the SDK client for a single call using the provided
+// token for per-rpc authentication. This should be used in request handlers where
+// Ensign requests are made on behalf of the user.
+func (c *EnsignClient) InvokeOnce(token string) *sdk.Client {
+	return c.client.WithCallOptions(auth.PerRPCToken(token, c.opts.Insecure))
```
pdeziel, line 35: I mostly added this to make the client calls easier to do  and to ensure that we're using the token in the handlers.

pkg/tenant/topics.go
bbengfort: There is - just a matter of getting to it!

bbengfort (approved):
Looks good; hopefully the clone method in the SDK doesn't cause a memory leak; but I think the way you're using it here means that we can just change the SDK (e.g. to return an interface) without having to worry about changing the code in tenant!

PR #473, opened 2023-05-04
Project patch routes

pdeziel:
### Scope of changes

This adds Tenant routes for patching projects. Currently, only the name, description, and owner can be patched.

Fixes SC-15840

### Type of change

- [x] new feature
- [ ] bug fix
- [ ] documentation
- [ ] testing
- [ ] technical debt
- [ ] other (describe)

### Acceptance criteria

### Author checklist

### Reviewer(s) checklist


pkg/tenant/projects.go
```diff
@@ -522,6 +568,155 @@ func (s *Server) ProjectUpdate(c *gin.Context) {
...
+				return
+			}
+
+			if description == project.Description {
+				c.JSON(http.StatusBadRequest, api.ErrorResponse("project already has this description"))
+				return
+			}
+
+			project.Description = description
+		case "owner":
```
pdeziel, line 672: Should this be `owner_id`? I don't like having to rely on hardcoded field names, is there something in the `confire` library that would help us parse the fields more elegantly?

pkg/tenant/projects.go
```diff
@@ -221,6 +221,52 @@ func (s *Server) TenantProjectCreate(c *gin.Context) {
...
+// Route: /tenant/:tenantID/projects/:projectID
+func (s *Server) TenantProjectPatch(c *gin.Context) {
+	var (
+		err                        error
+		orgID, tenantID, projectID ulid.ULID
+	)
+
+	// orgID is required to check ownership of the project
+	if orgID = orgIDFromContext(c); ulids.IsZero(orgID) {
+		return
```
bbengfort, line 236: Missing a `c.JSON` and/or sentry warning here?

pkg/tenant/projects.go
```diff
@@ -522,6 +568,155 @@ func (s *Server) ProjectUpdate(c *gin.Context) {
...
+// Route: /project/:projectID
+func (s *Server) ProjectPatch(c *gin.Context) {
+	var (
+		err              error
+		orgID, projectID ulid.ULID
+	)
+
+	// orgID is required to check ownership of the project
+	if orgID = orgIDFromContext(c); ulids.IsZero(orgID) {
+		return
```
bbengfort, line 583: Missing `c.JSON` and/or sentry warning here?

pkg/tenant/projects.go
```diff
@@ -221,6 +221,52 @@ func (s *Server) TenantProjectCreate(c *gin.Context) {
...
+	}
+
+	// Verify that the tenant exists in the user's organization.
+	if err = db.VerifyOrg(c, orgID, tenantID); err != nil {
+		if errors.Is(err, db.ErrNotFound) || errors.Is(err, db.ErrOrgNotVerified) {
+			c.JSON(http.StatusNotFound, api.ErrorResponse("tenant not found"))
+			return
+		}
+		sentry.Warn(c).Err(err).Msg("could not check verification")
+		c.JSON(http.StatusUnauthorized, api.ErrorResponse("could not verify organization"))
```
bbengfort, line 262: I think this should be a 500 error not a 401 -- most cases are captured by the not found above -- so this is probably a database error.

pkg/tenant/projects.go
```diff
@@ -522,6 +568,155 @@ func (s *Server) ProjectUpdate(c *gin.Context) {
...
+	}
+
+	// Verify that the project exists in the user's organization.
+	if err = db.VerifyOrg(c, orgID, projectID); err != nil {
+		if errors.Is(err, db.ErrNotFound) || errors.Is(err, db.ErrOrgNotVerified) {
+			c.JSON(http.StatusNotFound, api.ErrorResponse("project not found"))
+			return
+		}
+		sentry.Warn(c).Err(err).Msg("could not check verification")
+		c.JSON(http.StatusUnauthorized, api.ErrorResponse("could not verify organization"))
```
bbengfort, line 627: I think this should be a 500 error not a 401 -- most cases are captured by the not found above -- so this is probably a database error.

pkg/tenant/api/v1/api.go
```diff
@@ -39,13 +39,15 @@ type TenantClient interface {
 
 	TenantProjectList(ctx context.Context, id string, in *PageQuery) (*TenantProjectPage, error)
 	TenantProjectCreate(ctx context.Context, id string, in *Project) (*Project, error)
+	TenantProjectPatch(ctx context.Context, tenantID, projectID string, in map[string]interface{}) (*Project, error)
```
bbengfort, line 42: `in` can still be a `*Project` -- just make sure that there are `json:"omitempty"` tags in the `Project` struct so that the fields aren't present when they're sent to the server. None of the editable fields can be set to empty string or empty ID.

pkg/tenant/projects.go
bbengfort: It should match the field in the `Project` struct - so it's owner, but then you're expecting the value to be a `map[string]interface{}` that has the `id` field inside of it (e.g. the serialized `Member` struct`). 

Confire does have tools that will make `PATCH` easier, so we should create a follow on story to update it.

pkg/tenant/projects.go
```diff
@@ -522,6 +568,155 @@ func (s *Server) ProjectUpdate(c *gin.Context) {
...
+			var name string
+			if name, ok = value.(string); !ok {
+				c.JSON(http.StatusBadRequest, api.ErrorResponse(api.FieldTypeError(field, "string")))
+				return
+			}
+
+			if name == project.Name {
+				c.JSON(http.StatusBadRequest, api.ErrorResponse("project already has this name"))
+				return
+			}
```
bbengfort, line 656: I don't think a 400 is necessary in this case. PATCH shouldn't preclude the PUT case (e.g. the front-end might send an entire data representation for PATCH and that's fine).

pkg/tenant/projects.go
```diff
@@ -522,6 +568,155 @@ func (s *Server) ProjectUpdate(c *gin.Context) {
...
+	}
+
+	// Retrieve the project from the database.
+	if project, err = db.RetrieveProject(c, projectID); err != nil {
+		if errors.Is(err, db.ErrNotFound) {
+			c.JSON(http.StatusNotFound, api.ErrorResponse("project not found"))
+			return
+		}
+		sentry.Error(c).Err(err).Msg("could not retrieve project")
+		c.JSON(http.StatusInternalServerError, api.ErrorResponse("could not lookup project"))
```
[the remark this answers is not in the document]
pdeziel: Yeah that's probably a good idea.

pkg/tenant/projects.go
bbengfort: I suppose if a non-editable field is passed in we can return a 400.

pkg/tenant/projects.go
bbengfort: Thanks!

bbengfort (approved):
Adding a :shipit: since I think you know what needs to be done here.

PR #482, opened 2023-05-08
Sc 16865

daniellemaxwell:
### Scope of changes

Updates topic validation method to use TopicNameRegex. The following characters have been escaped within the expression: `.` `-` `_`. 

This also adds a length check to the topic name. (This has been added on the frontend, but was not previously done on the backend.)

Fixes SC-16865

### Type of change

- [ ] new feature
- [x] bug fix
- [ ] documentation
- [ ] testing
- [ ] technical debt
- [ ] other (describe)

### Acceptance criteria

### Author checklist

### Reviewer(s) checklist


pkg/tenant/db/errors.go
```diff
@@ -45,6 +45,7 @@ var (
 
 	// Max-length errors
 	ErrProjectDescriptionTooLong = errors.New("project description is too long")
+	ErrTopicNameTooLong          = errors.New("topic name is too long")
```
pdeziel, line 48: Can you make sure the frontend is validating this as well or create a story for it?

pkg/tenant/db/topics_test.go
```diff
@@ -72,8 +73,12 @@ func TestTopicValidate(t *testing.T) {
 	topic.Name = "otters;"
 	require.ErrorIs(t, topic.Validate(), db.ErrInvalidTopicName, "expected invalid name to be an error")
 
+	// Test name that is too long
+	topic.Name = strings.Repeat("a", 513)
+	require.ErrorIs(t, topic.Validate(), db.ErrTopicNameTooLong, "expected too long name to be an error")
```
pdeziel, line 78: Thanks for adding the test case!

pkg/tenant/db/topics_test.go
```diff
@@ -72,8 +73,12 @@ func TestTopicValidate(t *testing.T) {
...
+
 	// Valid topic
-	topic.Name = "otters"
+	topic.Name = "otters_are-cool"
```
pdeziel, line 81: Since the validation is a bit complicated, I think we should use a test case pattern similar to the following.

```golang
testCases := struct{}{
    name string,
    err error
}{
    ("otters;", db.ErrInvalidTopicName},
    ("otters_are-cool", nil),
}

for i, tc := testCases {
    if tc.err == nil {
        require.NoError(topic.Validate(), fmt.Sprintf("expected no error for test case %d", i)
    } else {
        require.ErrorIs(tc.err, topic.Validate(), fmt.Sprintf("expected error for test case %d", i)
    }
}
```

That will allows us to keep track of all the edge cases, so can you add a story for a future sprint to revise the test and add some more edge cases?

pdeziel (approved):
Looks great, thanks!

pkg/tenant/db/errors.go
daniellemaxwell: I'll double check the regex currently in use for validation on the FE side.

PR #545, opened 2023-06-20
Removes modified from api structs and responses

daniellemaxwell:
### Scope of changes

This removes `Modified` from the Tenant API structs since it is more for the database than API requests and responses.

Fixes SC-15633

### Type of change

- [ ] new feature
- [ ] bug fix
- [ ] documentation
- [ ] testing
- [x] technical debt
- [ ] other (describe)

### Acceptance criteria

[Copy acceptance criteria checklist from story for reviewer, or add a brief acceptance criteria here]

[Front-End: add screenshots/videos of changes made]

### Definition of Done

- [ ] I have manually tested the change running it locally (having rebuilt all containers) or via unit tests 

### Reviewer(s) checklist

### DIFF
--- a/pkg/tenant/api/v1/api.go
+++ b/pkg/tenant/api/v1/api.go
@@ -197,7 +197,6 @@ type Organization struct {
 	Projects  int    `json:"projects"`
 	LastLogin string `json:"last_login"`
 	Created   string `json:"created"`
-	Modified  string `json:"modified"`
 }
 
 type OrganizationPage struct {
@@ -210,7 +209,6 @@ type Tenant struct {
 	Name            string `json:"name"`
 	EnvironmentType string `json:"environment_type"`
 	Created         string `json:"created,omitempty"`
-	Modified        string `json:"modified,omitempty"`
 }
 
 type TenantPage struct {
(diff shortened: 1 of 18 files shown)


pkg/tenant/api/v1/api.go
```diff
@@ -324,7 +318,6 @@ type APIKey struct {
 	Owner        string   `json:"owner,omitempty"`
 	Permissions  []string `json:"permissions,omitempty"`
 	Created      string   `json:"created,omitempty"`
-	Modified     string   `json:"modified,omitempty"`
```
pdeziel, line 327: I think this is the only one I'm okay with removing. API keys are kind of a special case because they can't be edited. I believe every other resource can be edited, and we want to represent when that happened on the frontend.

pdeziel (commented):
I don't remember the initial motivation for this story, but it seems like we still want the modified timestamps on every resource except for API keys right?

pkg/tenant/api/v1/api.go
daniellemaxwell: @pdeziel The story originated from the following comment when timestamps were added to the `Member` struct : https://github.com/rotationalio/ensign/pull/359#discussion_r1153615902

The `Modified` timestamp isn't used in Beacon, so I thought removing from the other structs as well would work best as something like `LastUsed` or `LastActivity` would likely be preferred as it is in the `Member` struct. We can do whichever is preferred.

The comment linked to above mentions removing the `Created` timestamp from the `Member` struct as well and that could also be done since we're using `DateAdded`.

pdeziel: Gotcha, we can definitely rename the timestamps if that makes it easier for the frontend.

pdeziel (approved):
Makes sense to me!

PR #574, opened 2023-07-25
User token cache

pdeziel:
### Scope of changes

This adds a token cache utility which keeps track of tokens issued to users for projects and has LRU-like behavior to manage the cache size and expired tokens.

Fixes SC-18924

### Type of change

- [x] new feature
- [ ] bug fix
- [ ] documentation
- [ ] testing
- [ ] technical debt
- [ ] other (describe)

### Acceptance criteria

- Create LRU cache mapping user IDs to access tokens
- Cache should automatically expire entries where access token has expired
- Need to be able to call this from the Quarterdeck client or Tenant endpoints eventually

[Front-End: add screenshots/videos of changes made]

### Definition of Done

### Reviewer(s) checklist


pkg/quarterdeck/tokens/cache.go
```diff
@@ -0,0 +1,159 @@
...
+	return &Cache{
+		items:     list.New(),
+		index:     make(map[string]*list.Element, 0),
+		maxTokens: size,
+	}
+}
+
+// Get the token from the user ID and project ID or return an error if it does not
+// exist.
+func (c *Cache) Get(userID, projectID ulid.ULID) (tks string, err error) {
```
pdeziel, line 59: We need to store tokens by both user ID and project ID right? If the user is switching between projects, I can see a situation where the token from the previous project ID is loaded from the cache instead of the one they are trying to access.

bbengfort (approved):
:shipit: Went through the cache code, but jumping off the review for ensign office hours.

pkg/quarterdeck/tokens/cache.go
```diff
@@ -0,0 +1,159 @@
...
+	if token.expiresAt.Before(time.Now()) {
+		// If expired, remove the token from the cache and all tokens before it. This
+		// works because token expirations are monotonically increasing.
+		for element != nil {
+			prev := element.Prev()
+			c.remove(element)
+			element = prev
+		}
+
+		return "", ErrCacheExpired
```
bbengfort, line 89: Could we move this to its own function, e.g. `trunc` so we can benchmark it later?

pdeziel: Will do!

PR #592, opened 2023-07-31
Fix Quarterdeck Refresh Bug

bbengfort:
### Scope of changes

Fixes the Quarterdeck refresh bug that was caught in Sentry: could not retrieve user from database. It appears that what was happening was that the orgID was not on the refresh token claims, so the refresh ended up using an empty orgID for refresh so no token was returned. It is actually unclear if this will fix the issue, but instead of a 500 error being returned, a 400 error will be returned. 

Fixes SC-18236

### Type of change

- [ ] new feature
- [x] bug fix
- [ ] documentation
- [ ] testing
- [ ] technical debt
- [ ] other (describe)

### Acceptance criteria

Do we need to look deeper into PyEnsign to see if the problem is there? Should the default orgID have been returned? Is the issue that the subject of the access token is not correct?

### Definition of Done

- [ ] Check this box if a reviewer can merge this pull request after approval (leave it unchecked if you want to do it yourself)
- [x] Backend-end: Documented service configuration changes or created related devops stories

### Reviewer(s) checklist

### DIFF
--- a/pkg/quarterdeck/tokens/tokens.go
+++ b/pkg/quarterdeck/tokens/tokens.go
@@ -218,7 +218,8 @@ func (tm *TokenManager) CreateRefreshToken(accessToken *jwt.Token) (refreshToken
 			NotBefore: jwt.NewNumericDate(accessClaims.ExpiresAt.Add(tm.conf.RefreshOverlap)),
 			ExpiresAt: jwt.NewNumericDate(accessClaims.IssuedAt.Add(tm.conf.RefreshDuration)),
 		},
-		OrgID: accessClaims.OrgID,
+		OrgID:     accessClaims.OrgID,
+		ProjectID: accessClaims.ProjectID,
 	}
 	return tm.CreateToken(claims), nil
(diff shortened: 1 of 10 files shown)


pkg/quarterdeck/tokens/tokens.go
```diff
@@ -218,6 +218,7 @@ func (tm *TokenManager) CreateRefreshToken(accessToken *jwt.Token) (refreshToken
 			NotBefore: jwt.NewNumericDate(accessClaims.ExpiresAt.Add(tm.conf.RefreshOverlap)),
 			ExpiresAt: jwt.NewNumericDate(accessClaims.IssuedAt.Add(tm.conf.RefreshDuration)),
 		},
+		OrgID: accessClaims.OrgID,
```
bbengfort, line 221: This adds the `OrgID` to the refresh token claims so that they will be available during the `Refresh` process.

pdeziel (approved):
:shipit:

PR #601, opened 2023-08-01
Initial tenant query endpoint

pdeziel:
### Scope of changes

This adds an initial endpoint in Tenant for doing topic queries, currently hardcoded fixtures are used to unblock the frontend so there is more work to be done here to get this fully working.

Fixes SC-19466

### Type of change

- [x] new feature
- [ ] bug fix
- [ ] documentation
- [ ] testing
- [ ] technical debt
- [ ] other (describe)

### Acceptance criteria

This takes in the EnSQL query from the FE and returns maximum of 10 results in the response.
The result is either one of the standard errors (401 or 500, etc) or an error if the EnSQL query returns an error.

This needs to format the events for the response
(JSON - ensure proper spacing & indentation)
(Binary format - return base64 encoded version)

### Definition of Done

### Reviewer(s) checklist

### DIFF
--- a/pkg/tenant/api/v1/api.go
+++ b/pkg/tenant/api/v1/api.go
@@ -53,6 +53,8 @@ type TenantClient interface {
 	ProjectTopicList(ctx context.Context, id string, in *PageQuery) (*ProjectTopicPage, error)
 	ProjectTopicCreate(ctx context.Context, id string, in *Topic) (*Topic, error)
 
+	ProjectQuery(ctx context.Context, in *ProjectQueryRequest) (*ProjectQueryResponse, error)
+
 	TopicList(context.Context, *PageQuery) (*TopicPage, error)
 	TopicDetail(ctx context.Context, id string) (*Topic, error)
 	TopicStats(ctx context.Context, id string) ([]*StatValue, error)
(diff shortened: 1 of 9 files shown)


pkg/tenant/api/v1/api.go
```diff
@@ -279,6 +281,25 @@ type ProjectTopicPage struct {
...
+	Error   string         `json:"error,omitempty"`
+}
+
+type QueryResult struct {
+	Metadata        map[string]string `json:"metadata"`
+	Mimetype        string            `json:"mimetype"`
+	Version         string            `json:"version"`
+	IsBase64Encoded bool              `json:"is_base64_encoded"`
+	Data            string            `json:"data"`
+	Created         string            `json:"created"`
```
pdeziel, line 300: Let me know if I am missing any fields you are expecting here.

daniellemaxwell: In the Figma file [here](https://www.figma.com/file/PvL23fm6W36Zd88O1XVyIo/Beacon-DX-UI?type=design&node-id=8125-18524&mode=design&t=R8bRQrfWpjVBadm4-4), there's a # of results. Although, we'll only display 10, it appears that we want to show users the actual total even if it's more than 10. Should that be included here or handled elsewhere?

bbengfort: @pdeziel at the top level of the response (e.g. outside the list of results), return the total number of events the query would return. For now, it will just be the length of the list; but we can update it in the future when we have an `Explain` command for Ensign.

pdeziel: Thanks, makes sense to me.

PR #625, opened 2023-08-08
feat: [sc-19837] Create Staging Topics for QA

Aatmaj1:
Story details: https://app.shortcut.com/rotational/story/19837

### DIFF
--- a/scripts/publish_qa.py
+++ b/scripts/publish_qa.py
@@ -1,0 +1,39 @@
+import json
+import asyncio
+
+from pyensign.events import Event
+from pyensign.ensign import Ensign
+
+def load_event_fixtures(path):
+    events = []
+    print()
+    with open(path,'r') as f:
+        data = json.load(f)
+        for fixture in data:
+            events.append(Event(json.dumps(fixture["data"]).encode("utf-8"), mimetype=fixture["mimetype"]))
+    return events
+
+async def publish_fixtures(ensign_creds):
+    """
+    Read the JSON files from the fixtures directory and publish events to the test topics.
+    """
+    ensign = Ensign(
+        cred_path=ensign_creds,
+        endpoint="staging.ensign.world:443", 
+        auth_url="https://auth.ensign.world"
+        )
+
+    # A topic that contains only one mimetype
+    # Less than 10 events
+    events = load_event_fixtures('fixtures/one_type.json')
+    for event in events:
+        await ensign.publish("documents_one_type", event)
+
+    # A topic that contains multiple mimetypes
+    # More than 10 events - we want to test that only 10 events are returned when the FE queries the topic
+    events = load_event_fixtures('fixtures/two_type.json')
+    for event in events:
+        await ensign.publish("documents_two_type", event)
+
+if __name__ == "__main__":
+    asyncio.run(publish_fixtures(ensign_creds='secret/qa_creds.json'))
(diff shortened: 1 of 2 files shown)


scripts/publish_qa.py
```diff
@@ -0,0 +1,39 @@
...
+from pyensign.events import Event
+from pyensign.ensign import Ensign
+
+def load_event_fixtures(path):
+    events = []
+    print()
+    with open(path,'r') as f:
+        data = json.load(f)
+        for fixture in data:
+            events.append(Event(json.dumps(fixture["data"]).encode("utf-8"), mimetype=fixture["mimetype"]))
```
pdeziel, line 13: We will also want to include the `type` (name and version), however it's not possible to do so with the current version of PyEnsign.

pdeziel (commented):
Looks good so far, can you also push the fixture files?

scripts/fixtures_qa/one_type.json
```diff
@@ -0,0 +1,119 @@
+[
+        {
+            "mimetype": 50,
+            "type": {
+                "name": "Generic",
+                "major_version": 1, 
+                "minor_version": 0, 
+                "patch_version": 0
+            },
```
pdeziel, line 9: Instead of the `type` field, in this file can we add two fields called `schema_name` and `schema_version`? e.g.

```
"schema_name": "Trade",
"schema_version": "1.0.0"
```

scripts/fixtures_qa/one_type.json
```diff
@@ -0,0 +1,119 @@
...
+            "data": {
+                "price": 23.09, "symbol": "AMZN", "timestamp": 1690928634000, "volume": 300
+            }
```
pdeziel, line 12: Let's also add a `meta` field with key-value strings. They don't have to be the same for each event, but maybe you can do something like:

```
"meta": {
     "symbol": "AMZN",
}
```

Just so we have something to display for metadata on the frontend side.

scripts/fixtures_qa/one_type.json
pdeziel: This will be more compatible with the upcoming version of PyEnsign.

scripts/fixtures_qa/two_type.json
```diff
@@ -0,0 +1,213 @@
+[
+    {
+        "mimetype": 50, 
+        "type": {
+            "name": "Generic",
+            "major_version": 1, 
+            "minor_version": 0, 
+            "patch_version": 0
+        },
```
pdeziel, line 9: Similarly in this file, let's change the `type` to `schema_name` and `schema_version`, although in this file you should create two separate type names/versions for the trades events and the document events.

scripts/fixtures_qa/two_type.json
```diff
@@ -0,0 +1,213 @@
...
+                "last_modified":"",
+                "active": "True",
+                "status_code":0,
+                "error":"",
+                "fetched_at": "ExtType(code=5, data='')",
+                "feed_id":"",
+                "language":"",
+                "year":"",
+                "month":"",
+                "day":0,
```
pdeziel, line 68: Did we lose data here somehow?

pdeziel (commented):
Thanks for adding the files, could you make some changes to the fixtures as I've requested here? I'm just trying to make the data a bit more meaningful for QA.

PR #639, opened 2023-08-10
Integrate query endpoint with ensign

pdeziel:
### Scope of changes

This makes the Tenant query endpoint functional by integrating the call to the Ensign query RPC.

Fixes SC-19587

### Type of change

- [x] new feature
- [ ] bug fix
- [ ] documentation
- [ ] testing
- [ ] technical debt
- [ ] other (describe)

### Acceptance criteria

This RPC takes an EnSQL query and returns a response with the query results which are some representation of the events
This endpoint should be authenticated with the appropriate permissions

### Definition of Done

### Reviewer(s) checklist

### DIFF
--- a/pkg/tenant/api/v1/api.go
+++ b/pkg/tenant/api/v1/api.go
@@ -57,6 +57,7 @@ type TenantClient interface {
 
 	TopicList(context.Context, *PageQuery) (*TopicPage, error)
 	TopicDetail(ctx context.Context, id string) (*Topic, error)
+	TopicEvents(ctx context.Context, id string) ([]*EventTypeInfo, error)
 	TopicStats(ctx context.Context, id string) ([]*StatValue, error)
 	TopicUpdate(context.Context, *Topic) (*Topic, error)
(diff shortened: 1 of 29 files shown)


pkg/tenant/api/v1/api.go
```diff
@@ -283,8 +283,14 @@ type ProjectTopicPage struct {
 }
 
 type ProjectQueryRequest struct {
-	ProjectID string `json:"project_id"`
-	Query     string `json:"query"`
+	ProjectID  string            `json:"project_id"`
+	Query      string            `json:"query"`
+	Parameters []*QueryParameter `json:"parameters"`
```
pdeziel, line 288: Added the parameters as a placeholder for right now, I would need to think more about how to deal with the parameter types.

bbengfort: I think parameters are most useful for the SDK where users are trying to sanitize user input -- we will probably not use parameters from Beacon, so it's not really something Tenant has to worry about. You don't have to remove the parameters if you want to leave the stubs in.

bbengfort (approved):
Looks good - thanks for getting that in so fast!

PR #674, opened 2023-08-30
Onboarding validation

pdeziel:
### Scope of changes

This updates the member validation to include validating the onboarding fields, which potentially returns multiple errors.

Fixes SC-20543

### Type of change

- [x] new feature
- [ ] bug fix
- [ ] documentation
- [ ] testing
- [ ] technical debt
- [ ] other (describe)

### Acceptance criteria

- All non-empty fields should be validated, if the field is empty then it's not validated.
- Should only return True/No errors if all fields are provided and valid (there are no optional fields)
- Multiple errors may be returned, the backend won't know anything about onboarding order or what step the user is on

### Definition of Done

### Reviewer(s) checklist

### DIFF
--- a/pkg/tenant/db/errors.go
+++ b/pkg/tenant/db/errors.go
@@ -3,6 +3,7 @@ package db
 import (
 	"errors"
 	"fmt"
+	"strings"
 )
 
 var (
@@ -18,7 +19,6 @@ var (
 	ErrMissingTenantID     = errors.New("object requires tenant id for serialization")
 	ErrMissingProjectID    = errors.New("object requires project id for serialization")
 	ErrMissingMemberEmail  = errors.New("member email is required")
-	ErrMissingMemberName   = errors.New("member name is required")
 	ErrMissingMemberRole   = errors.New("member role is required")
 	ErrMissingMemberStatus = errors.New("member status is required")
 	ErrMissingProjectName  = errors.New("project name is required")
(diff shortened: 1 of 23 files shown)


pkg/tenant/db/errors.go
```diff
@@ -49,22 +54,44 @@ var (
 	ErrTopicNameTooLong          = errors.New("topic name is too long")
 )
 
+// ValidationError represents a validation error for a specific field and may contain
+// an index if the field is a slice.
 type ValidationError struct {
-	err error
+	Field string
+	Err   error
+	Index int
```
bbengfort, line 62: This data structure will have to be JSON serialized for the front end:

```suggestion
	Field string `json:"field"`
	Err   error `json:"error"
	Index int `json:"index"
```

pkg/tenant/db/errors.go
```diff
@@ -49,22 +54,44 @@ var (
...
+}
+
+func (e *ValidationError) AtIndex(index int) *ValidationError {
+	e.Index = index
+	return e
 }
 
 func (e *ValidationError) Error() string {
-	return fmt.Sprintf("validation error: %s", e.err)
+	return fmt.Sprintf("invalid field '%s': %s", e.Field, e.Err)
```
bbengfort, line 78: Should we handle the case where there is no field or the case where there is an index in the string?

pkg/tenant/db/errors.go
```diff
@@ -49,22 +54,44 @@ var (
...
+}
+
+type ValidationErrors []*ValidationError
+
+func (v ValidationErrors) Error() string {
+	errs := make([]string, 0, len(v))
+	for _, e := range v {
+		errs = append(errs, e.Error())
+	}
+	return fmt.Sprintf("%d validation errors occurred:\n%s", len(v), strings.Join(errs, "\n"))
```
bbengfort, line 96:
```suggestion
	return fmt.Sprintf("%d validation errors occurred:\n  -%s", len(v), strings.Join(errs, "\n  -"))
```

A bit nicer formatting suggestion

bbengfort (approved):
Nice work! I like the new `ValidationError` data structure. My only concern is that the ValidationError needs to be serialized as JSON in the response; does `api.ErrorResponse` handle this type correctly?

pkg/tenant/db/errors.go
pdeziel: I was thinking of adding a similar struct to the API package, except that it contains the JSON tags for serialization, and then converting the one here into the API response struct. But if that's too clunky we could just add the tags here.

bbengfort: That makes sense to me - whatever you'd prefer; just making sure I wasn't missing something in translation!

PR #679, opened 2023-09-01
Update member profile endpoint to support onboarding

pdeziel:
### Scope of changes

This updates the member profile update endpoint to be compatible with the new onboarding validation, and adds a Quarterdeck endpoint to support organization updates.

Fixes SC-20554

### Type of change

- [x] new feature
- [ ] bug fix
- [ ] documentation
- [ ] testing
- [ ] technical debt
- [ ] other (describe)

### Acceptance criteria

- Should be an authenticated endpoint that takes in an onboarding form (basically the member profile).
- If there are validation errors on provided fields, those errors should be returned. Empty fields are not validated.
- If all fields are provided and valid, return completed status (and record the onboarded state in the backend)
- Need to handle the is_invited state for the user (e.g. organization name is not editable)
- Updates should be made to Quarterdeck (org name and workspace), but only if not invited

### Definition of Done

### Reviewer(s) checklist


pkg/tenant/members.go
```diff
@@ -238,66 +242,94 @@ func (s *Server) MemberUpdate(c *gin.Context) {
...
+	member.ProfessionSegment = req.ProfessionSegment
+	member.DeveloperSegment = req.DeveloperSegment
+
+	if !member.Invited {
+		member.Organization = req.Organization
+		member.Workspace = req.Workspace
+	}
+
+	// Validate the member update
+	if err = member.Validate(); err != nil {
```
pdeziel, line 281: Okay this is a bit awkward because we're validating the database record twice (here and in `db.UpdateMember`), but I wanted to make sure that the organization and workspace are valid before sending it to Quarterdeck.

pkg/tenant/topics.go
```diff
@@ -369,6 +369,7 @@ func (s *Server) TopicEvents(c *gin.Context) {
 	if topicID, err = ulid.Parse(c.Param("topicID")); err != nil {
 		sentry.Warn(c).Err(err).Str("topicID", c.Param("topicID")).Msg("could not parse topic id")
 		c.JSON(http.StatusNotFound, api.ErrorResponse(responses.ErrTopicNotFound))
+		return
```
pdeziel, line 372: Not sure why this was not causing a test failure before.

pkg/quarterdeck/mock/quarterdeck.go
```diff
@@ -390,6 +394,10 @@ func (s *Server) OrganizationsCount(param string) int {
 	return s.count(methodPath(http.MethodGet, fullPath(OrganizationsEP, param)))
 }
 
+func (s *Server) OrganizationsUpdateCount(id string) int {
+	return s.count(methodPath(http.MethodPut, fullPath(OrganizationsEP, id)))
+}
+
```
bbengfort, line 400: Thanks for adding this in as well.

pkg/tenant/api/v1/api.go
```diff
@@ -240,6 +241,21 @@ type Member struct {
 	LastActivity      string   `json:"last_activity,omitempty"`
 }
 
+// Normalize performs some cleanup on the Member fields to ensure that fields provided
+// in the JSON request can be used in comparisons and uniqueness checks.
+func (m *Member) Normalize() {
+	m.Email = strings.TrimSpace(m.Email)
```
bbengfort, line 247: should we also lower case the email?

pkg/tenant/api/v1/api.go
```diff
@@ -240,6 +241,21 @@ type Member struct {
...
+	m.Name = strings.TrimSpace(m.Name)
+	m.Organization = strings.TrimSpace(m.Organization)
+	m.Workspace = strings.ToLower(strings.TrimSpace(m.Workspace))
+	m.ProfessionSegment = strings.TrimSpace(m.ProfessionSegment)
+	for i, s := range m.DeveloperSegment {
+		m.DeveloperSegment[i] = strings.TrimSpace(s)
```
bbengfort, line 253: Do we have an enum for the developer segments?

pkg/tenant/api/v1/client.go
```diff
@@ -1073,13 +1073,27 @@ func (s *APIv1) Do(req *http.Request, data interface{}, checkStatus bool) (rep *
...
 				}
 			}
+
+			// Otherwise attempt to parse as a FieldValidationErrors
+			var verrs FieldValidationErrors
+			if err = json.Unmarshal(bytes, &verrs); err == nil {
+				if len(verrs) > 0 {
+					return rep, fmt.Errorf("[%d] %s", rep.StatusCode, verrs.Error())
+				}
+			}
```
bbengfort, line 1095: Wouldn't validation errors have a `400` error code?

pkg/tenant/api/v1/errors.go
```diff
@@ -28,6 +29,36 @@ var (
...
+	case e.Field == "":
+		return e.Err
+	case e.Index > -1:
+		return fmt.Sprintf("%q at index %d: %s", e.Field, e.Index, e.Err)
+	default:
+		return fmt.Sprintf("%q: %s", e.Field, e.Err)
+	}
+}
+
+type FieldValidationErrors []*FieldValidationError
```
bbengfort, line 52: I was thinking we could add this to the standard `Reply` struct for 400 errors and have `json:"omitempty"` for this and the `Error` field so that it would be one or the other in responses.

pkg/tenant/db/errors.go
```diff
@@ -103,3 +105,15 @@ func (v ValidationErrors) Error() string {
 	}
 	return fmt.Sprintf("%d validation errors occurred:\n  -%s", len(v), strings.Join(errs, "\n  -"))
 }
+
+func (v ValidationErrors) ToAPI() api.FieldValidationErrors {
```
bbengfort, line 109: Nice.

pkg/tenant/db/members.go
```diff
@@ -26,7 +26,7 @@ const (
 
 var (
 	// Must be at least 3 characters, cannot start with a number, and is alphanumeric with + _ and -
-	WorkspaceNameRegex = regexp.MustCompile(`^[a-zA-Z][a-zA-Z0-9+_-]{2,}$`)
+	WorkspaceNameRegex = regexp.MustCompile(`^[a-zA-Z][a-zA-Z0-9.+_-]{2,}$`)
```
bbengfort, line 29: what prompted this? Are `.` allowed in slugs?

bbengfort: I'm not sure we can allow `.` if we're going to use this for subdomains in the future.

pkg/tenant/members.go
bbengfort: I'm not bothered by double validation - but I would recommend that you add that note to the comment so we remember why we're doing it twice in the future.

pkg/tenant/topics.go
bbengfort: Whoops, good catch. I wonder how we can test for this …

bbengfort (approved):

pkg/tenant/api/v1/api.go
pdeziel: We don't have one currently, should I create a story for that and the profession segment?

pkg/tenant/db/members.go
pdeziel: My bad, somehow I thought I'd just overlooked this in the last PR.

pkg/tenant/api/v1/errors.go
pdeziel: Yes, that makes sense to me.

pkg/tenant/api/v1/api.go
bbengfort: Yes please, I think that's a good idea.

PR #681, opened 2023-09-01
Quarterdeck accept invite

pdeziel:
### Scope of changes

This adds an authenticated endpoint to Quarterdeck for accepting user invites and refactors some logic into a shared helper method.

Fixes SC-20702

### Type of change

- [x] new feature
- [ ] bug fix
- [ ] documentation
- [ ] testing
- [ ] technical debt
- [ ] other (describe)

### Acceptance criteria

Port the accept invitation code from the login endpoint into an accept invitation endpoint. On invalid invitation this should return 400 and on valid invitation the user should be put on the organization.

### Definition of Done

### Reviewer(s) checklist

### DIFF
--- a/pkg/quarterdeck/api/v1/client.go
+++ b/pkg/quarterdeck/api/v1/client.go
@@ -489,11 +489,9 @@ func (s *APIv1) InviteCreate(ctx context.Context, in *UserInviteRequest) (out *U
 	return out, nil
 }
 
-func (s *APIv1) InviteAccept(ctx context.Context, token string) (out *LoginReply, err error) {
-	endpoint := fmt.Sprintf("/v1/invites/%s", token)
-
+func (s *APIv1) InviteAccept(ctx context.Context, in *UserInviteToken) (out *LoginReply, err error) {
 	var req *http.Request
-	if req, err = s.NewRequest(ctx, http.MethodPost, endpoint, nil, nil); err != nil {
+	if req, err = s.NewRequest(ctx, http.MethodPost, "/v1/invites/accept", in, nil); err != nil {
 		return nil, err
 	}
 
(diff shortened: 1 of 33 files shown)


pkg/quarterdeck/invites.go
```diff
@@ -198,3 +202,140 @@ func (s *Server) InviteCreate(c *gin.Context) {
...
+		c.JSON(http.StatusInternalServerError, api.ErrorResponse(responses.ErrSomethingWentWrong))
+		return err
+	}
+
+	// Verify that the invite is for this email address
+	if err = invite.Validate(user.Email); err != nil {
+		switch {
+		case errors.Is(err, db.ErrTokenExpired):
+			log.Debug().Msg("invite token expired")
+			metrics.FailedLogins.WithLabelValues(ServiceName, UserHuman, "invite token expired").Inc()
```
pdeziel, line 302: Not sure if this is going to break the metrics, these strings are a bit different than they were in Login.

bbengfort (approved):
This was very tough to review -- I didn't see anything concerning but I think it would be good if you, me, and @pdamodaran could step through each line of code to make sure we're all on the same page with what's happening during office hours?

pkg/quarterdeck/api/v1/client.go
```diff
@@ -485,6 +489,21 @@ func (s *APIv1) InviteCreate(ctx context.Context, in *UserInviteRequest) (out *U
 	return out, nil
 }
 
+func (s *APIv1) InviteAccept(ctx context.Context, token string) (out *LoginReply, err error) {
+	endpoint := fmt.Sprintf("/v1/invites/%s", token)
```
bbengfort, line 493: It's not typical that you POST `nil` data to an endpoint; the token is data not an identity. Is it possible to do `POST /v1/invites` and put the token in the body?

bbengfort: Or alternatively do a POST to `/v1/invites/:userID`? The client could then parse the userID out of the token.

PR #684, opened 2023-09-04
Tenant accept invite

pdeziel:
### Scope of changes

This adds an endpoint to Tenant for accepting user invites, which is a forwarded request to Quarterdeck.

Fixes SC-20703

### Type of change

- [x] new feature
- [ ] bug fix
- [ ] documentation
- [ ] testing
- [ ] technical debt
- [ ] other (describe)

### Acceptance criteria

This endpoint should call the Quarterdeck accept invite endpoint, if successful then the member record should be updated and credentials should be returned to the caller.

### Definition of Done

### Reviewer(s) checklist


pkg/tenant/api/v1/client.go
```diff
@@ -52,6 +53,7 @@ type APIv1 struct {
 	endpoint *url.URL
 	client   *http.Client
 	creds    string
+	cookies  []*http.Cookie
```
pdeziel, line 56: I wasn't sure how to capture the session cookies for the tests (I think we need to be using https to have access to those?) so this is a workaround to ensure that cookies are set. Is there a better way to do this?

pkg/tenant/invites.go
```diff
@@ -0,0 +1,149 @@
...
+	var userID ulid.ULID
+	if userID = newClaims.ParseUserID(); ulids.IsZero(userID) {
+		sentry.Error(c).Msg("could not parse userID from newly issued claims")
+		c.JSON(http.StatusInternalServerError, api.ErrorResponse(responses.ErrSomethingWentWrong))
+		return
+	}
+
+	// Retrieve member from the database to update the invitation status
+	var member *db.Member
+	if member, err = db.RetrieveMember(ctx, orgID, userID); err != nil {
```
pdeziel, line 125: So if the member record is not in the database and we error here, it might create a situation where the user is in the Quarterdeck org but not in trtl. In that case, should we create a new member here or deal with it manually?

pkg/tenant/members.go
```diff
@@ -560,40 +560,6 @@ func (s *Server) MemberDelete(c *gin.Context) {
...
-// endpoint must not be authenticated because unauthorized users should be able to
-// accept organization invitations. Frontends should use this endpoint to validate an
-// invitation token after the user has clicked on an invitation link in their email.
-// The preview must contain enough information so the user knows which organization
-// they are joining and also whether or not the email address is already registered to
-// an account. This allows frontends to know whether or not to prompt the user to
-// login or to create a new account.
-//
-// Route: /invites/:token
-func (s *Server) InvitePreview(c *gin.Context) {
```
pdeziel, line 573: Moved to `invites.go`

pkg/tenant/api/v1/client.go
```diff
@@ -1153,3 +1166,49 @@ func (c *APIv1) SetCSRFProtect(protect bool) error {
...
+
+// Returns the cookies set from the previous request(s) on the client Jar.
+func (c *APIv1) Cookies() (_ []*http.Cookie, err error) {
+	if c.client.Jar == nil {
+		return nil, ErrNoCookies
+	}
+
+	cookies := c.client.Jar.Cookies(c.endpoint)
+	return cookies, nil
+}
```
bbengfort, line 1214: Per your question above, I did try to add access and refresh token handlers for TLS. Not quite working yet, but I made a start.

pkg/tenant/invites_test.go
```diff
@@ -0,0 +1,177 @@
...
+	require.NoError(s.SetClientCredentials(claims), "could not set client credentials")
+	err = s.client.InviteAccept(ctx, req)
+	s.requireError(err, http.StatusBadRequest, responses.ErrTryLoginAgain, "expected error when token is missing")
+
+	// Successful invite accept request
+	req.Token = "token1234"
+	err = s.client.InviteAccept(ctx, req)
+	require.NoError(err, "could not accept invite")
+	token, err := s.GetClientAccessToken()
+	require.NoError(err, "could not get access cookie token")
```
bbengfort, line 160: This is failing right now, I'm trying to figure out why.

pkg/tenant/tenant_test.go
```diff
@@ -112,6 +115,10 @@ func (suite *tenantTestSuite) SetupSuite() {
 	suite.srv, err = tenant.New(conf)
 	assert.NoError(err, "could not create the tenant api server from the test configuration")
 
+	suite.tlssrv = httptest.NewUnstartedServer(nil)
+	suite.srv.Server = *service.New(conf.BindAddr, service.WithMode(conf.Mode), service.WithTestServer(suite.tlssrv))
+	suite.srv.Server.Register(suite.srv)
```
bbengfort, line 120: This is where I attempted to convert the server into a TLS server so that the cookies would be set correctly; I think I'm 90% of the way there, just missing one thing but I can't figure out what.

bbengfort: Ah, I think I know - I'm not running `ServeTLS` … let me see if that fixes it.

pkg/utils/tlstest/testcert.go
bbengfort, line 1: Had to port this file from the Go standard library since it was `internal` and therefore unable to be imported directly.

pkg/utils/tlstest/tlstest.go
bbengfort, line 1: This file extracts the basic parts out of the `httptest.StartTLS` method:

https://cs.opensource.google/go/go/+/refs/tags/go1.21.1:src/net/http/httptest/server.go;l=142-186

pkg/tenant/invites.go
bbengfort: For now, let's deal with it manually; the way that would happen is if something in the invitation process went wrong, right? I'd rather know about upstream errors for right now.

pkg/tenant/invites_test.go
bbengfort: @pdeziel ok, here is why the tests are failing: the max age on the cookies is 0 because the max age is computed from the access and refresh tokens expiration. When the max age is 0 or negative, the cookie jar deletes the cookie instead of setting it. Instead of using constant token strings, we need to use strings that have an expiration in the future for the test to pass.

bbengfort (approved):
Ok, sorry about all the changes, but I think I sorted out the cookie stuff!

PR #714, opened 2023-09-15
Sc 20587

daniellemaxwell:
### Scope of changes

Onboarding e2e test for new users.

Fixes SC-20587

### Type of change

- [ ] new feature
- [ ] bug fix
- [ ] documentation
- [x] testing
- [ ] technical debt
- [ ] other (describe)

### Acceptance criteria

Screenshot 2023-09-15 at 3 37 04 PM (2)

### Definition of Done

- [ ] I have manually tested the change running it locally (having rebuilt all containers) or via unit tests 
- [ ] Check this box if a reviewer can merge this pull request after approval (leave it unchecked if you want to do it yourself)

### Reviewer(s) checklist


web/beacon-app/src/features/onboarding/hooks/useWorkspaceForm.tsx
```diff
@@ -13,7 +13,8 @@ export const FORM_INITIAL_VALUES = {
 export const FORM_VALIDATION_SCHEMA = object({
   workspace: string()
     .trim()
-    .required(t`Workspace name is required.`),
+    .required(t`Workspace name is required.`)
+    .matches(/^[a-zA-Z][a-zA-Z0-9-_]{2,}$/, t`Workspace must have at least 3 characters and cannot begin with a number.`),
```
daniellemaxwell, line 17: Before adding this to the validation schema, users would only see a message that the workspace they entered already existed. This should provide a more helpful error message if their workspace is too short or begins with a number.

masskoder: Looks good. However, if the backend validation rules change in the future, we may deliver incorrect or inconsistent messages to the user. so we will have the same issue again since in the backend all 400 error code is related to already existing value.

https://github.com/rotationalio/ensign/blob/eb9081b1ea17689c7e6de065b8c3042c2d2ed922/web/beacon-app/src/features/onboarding/components/steps/workspace/index.tsx#L17

Frontend validation is very great, but in this case, it would be ideal to also identify the type of error based on backend feedback.
@pdeziel Would it be possible to add specific error codes that differentiate between validation issues and errors related to already existing values? Would it be better to use different HTTP response codes for these scenarios?


Could be done in the future but it's good to go.

masskoder (approved):
LGTM

web/beacon-app/src/features/onboarding/hooks/useWorkspaceForm.tsx
masskoder: Looks good. However, if the backend validation rules change in the future, we may face the issue of delivering incorrect or inconsistent messages to the user. 


While frontend validation is useful, it would be ideal to also identify the type of error based on backend feedback. @pdeziel, could we consider adding specific error codes to distinguish between validation issues and errors related to duplicate values? Alternatively, should we use different HTTP response codes for these scenarios?

This could be done in the future.

pdeziel: @masskoder The backend should be returning 409 for the already exists case, is that not happening?

daniellemaxwell: @masskoder @pdeziel I think the issue is that the check may not be looking for the correct error code. I'll look into it and update.

masskoder: @daniellemaxwell Sure, I created a story for that, if you can use it to refactor this, it would be great.

PR #739, opened 2023-09-25
Event Hash Methods

bbengfort:
### Scope of changes

Implements event hashing methods for deduplication and duplicate detection.

Fixes SC-20991

### Type of change

- [x] new feature
- [ ] bug fix
- [ ] documentation
- [ ] testing
- [ ] technical debt
- [ ] other (describe)

### Acceptance criteria

There is enough test coverage to make us confident that these hashing mechanisms will catch duplicates efficiently. 

### Definition of Done

- [x] I have recompiled and included new protocol buffers to reflect changes I made if necessary
- [x] Backend-end: Documented service configuration changes or created related devops stories

### Reviewer(s) checklist


proto/api/v1beta1/topic.proto
```diff
@@ -112,9 +112,10 @@ message Deduplication {
         UNKNOWN = 0;
         NONE = 1;
         STRICT = 2;
-        KEY_GROUPED = 3;
-        UNIQUE_KEY = 4;
-        UNIQUE_FIELD = 5;
+        DATAGRAM = 3;
+        KEY_GROUPED = 4;
+        UNIQUE_KEY = 5;
+        UNIQUE_FIELD = 6;
```
bbengfort, line 118: This change shouldn't be a problem since this is not deployed yet.

go.mod
```diff
@@ -29,6 +29,7 @@ require (
 	github.com/urfave/cli/v2 v2.25.7
 	github.com/vmihailenco/msgpack/v5 v5.3.5
 	golang.org/x/crypto v0.13.0
+	golang.org/x/exp v0.0.0-20230905200255-921286631fa9
```
bbengfort, line 32: using the experimental `slices` package for sorting; once we upgrade to Go 1.21 -- this will be part of stdlib and go away.

pkg/ensign/api/v1beta1/deduplication.go
```diff
@@ -0,0 +1,194 @@
...
+	default:
+		return nil, fmt.Errorf("unknown deduplication strategy %s", policy.Strategy)
+	}
+}
+
+// Strict hashing is used to detect duplicates where two events have identical metadata,
+// data, mimetype, and type. This method works by setting any non-hash fields to zero
+// values then marshaling the protocol buffers of the event and computing the murmur3
+// hash on the serialized data.
+func (w *EventWrapper) HashStrict() (_ []byte, err error) {
```
bbengfort, line 47: It turns out that `proto.Marshal` is not deterministic, and even though there is a setting for marshal deterministic, it is not stable between languages or builds. As a result, I've had to manually serialize the event struct myself to create the hash for strict.

pkg/ensign/api/v1beta1/deduplication.go
```diff
@@ -0,0 +1,194 @@
...
+	if _, err = hash.Write(event.Data); err != nil {
+		return nil, fmt.Errorf("could not write event data to hash: %w", err)
+	}
+
+	// Sort the keys in the metadata to write them in lexicographic order
+	keys := make([]string, 0, len(event.Metadata))
+	for key := range event.Metadata {
+		keys = append(keys, key)
+	}
+	slices.Sort(keys)
```
bbengfort, line 64: Sorting the keys is essential, key order is not necessarily deterministic in Go if you range over the map; so by guaranteeing a specific order, the hash remains stable.

pkg/ensign/api/v1beta1/deduplication.go
```diff
@@ -0,0 +1,194 @@
...
+
+	// Write the metadata to the to the hash
+	for _, key := range keys {
+		val := event.Metadata[key]
+		if _, err = hash.Write([]byte(key + val)); err != nil {
```
bbengfort, line 69: Including the key and the value in the meta data hash, concatenated with no separator.

pkg/ensign/api/v1beta1/deduplication.go
```diff
@@ -0,0 +1,194 @@
...
+			return nil, fmt.Errorf("could not write metadata key value pair to hash: %w", err)
+		}
+	}
+
+	// Write the mimetype to the hash
+	if err = binary.Write(hash, binary.LittleEndian, event.Mimetype); err != nil {
```
bbengfort, line 75: All integers are written using LittleEndian encoding.

pkg/ensign/api/v1beta1/deduplication.go
```diff
@@ -0,0 +1,194 @@
...
+		return nil, err
+	}
+
+	hash := murmur3.New128()
+	for _, key := range keys {
+		val := event.Metadata[key]
+		if _, err = hash.Write([]byte(key + val)); err != nil {
+			return nil, err
+		}
+	}
```
bbengfort, line 146: the order of the keys specified to this function will determine the order of the key value pairs in the hash. I'm also concatenating the key + value in the hash but I'm not sure this is necessary. E.g. what if the event has none of the keys specified by the policy, we'd just be writing the key names to the event. This is a good thing if the hash should not be stable wrt policy changes (e.g. adding or removing keys) … but I'm not sure it's adding anything to the code.

pkg/ensign/api/v1beta1/deduplication.go
```diff
@@ -0,0 +1,194 @@
...
```
bbengfort, line 174: See above comment about hashing key/value, policy order, and whether or not the concatenation is necessary.

pkg/ensign/api/v1beta1/deduplication.go
```diff
@@ -0,0 +1,194 @@
...
+
+	return hash.Sum(nil), nil
+}
+
+// Unique field hashing determines duplicates not from the entire datagram, but rather
+// from specified fields in the datagram. This requires Ensign to be able to parse the
+// data, and unparsable mimetypes (such as protocol buffers) will return an error.
+//
+// BUG: this is currently unimplemented
+func (w *EventWrapper) HashUniqueField(fields []string) (_ []byte, err error) {
```
bbengfort, line 184: To get a unique field we need to be able to unmarshal the data …

pdeziel: Ok so this is blocked by not having an event parser yet?

pkg/ensign/api/v1beta1/deduplication_test.go
```diff
@@ -0,0 +1,510 @@
...
+	testCases := []struct {
+		event    *EventWrapper
+		expected []byte
+		err      error
+	}{
+		{
+			&EventWrapper{Event: nil}, nil, ErrNoEvent,
+		},
+		{
+			fixtures[0], []byte{0xb, 0xa6, 0x75, 0xb8, 0x91, 0xd7, 0xee, 0xc8, 0xa4, 0xe1, 0xf, 0x84, 0x7d, 0x8a, 0x14, 0x9b}, nil,
```
pdeziel, line 62: How were the expected values generated?

pkg/ensign/api/v1beta1/deduplication.go
pdeziel: I see your point, I wonder if we could make that configurable? So it can either be unique by keys or also by key+value?

pdeziel (approved):
Looks good, and looks like there's a good start on the test coverage.

pkg/ensign/api/v1beta1/deduplication.go
pdamodaran: I think concatenating the key + value in the hash works for bookkeeping purposes, we know which keys were used for hashing.  This is also helpful for the reason you pointed out, that the hash should not be stable wrt policy changes.

pdamodaran (approved):
Looks good @bbengfort!  I am looking forward to trying this out in Eisenhower once it is deployed.

pkg/ensign/api/v1beta1/deduplication.go
bbengfort: Ok - thank you for your input!

pkg/ensign/api/v1beta1/deduplication.go
bbengfort: That's correct - we need some way to look into the data in order to select fields.

PR #747, opened 2023-09-27
Reset password token

pdeziel:
### Scope of changes

This adds a password reset token struct and methods to sign/verify the token.

Fixes SC-20879

### Type of change

- [x] new feature
- [ ] bug fix
- [ ] documentation
- [ ] testing
- [ ] technical debt
- [ ] other (describe)

### Acceptance criteria

Create a data structure that can be signed and serialized/deserialized to use as a token to send in password reset emails.

Data included:

User ID
Expiration
Random nonce
Key ID
Default the expiration to 15 minutes from now when creating the data struct.

Create marshal/unmarshal methods for converting the token into a signed string. Create validate method for validating that the token was signed by Quarterdeck. Write tests for these methods.

### Definition of Done

### Reviewer(s) checklist


pkg/quarterdeck/db/tokens.go
```diff
@@ -57,29 +52,76 @@ func (t *VerificationToken) Sign() (_ string, secret []byte, err error) {
...
+
+	if _, err = rand.Read(token.Nonce); err != nil {
+		return nil, fmt.Errorf("could not generate token: %w", err)
+	}
+	return token, nil
+}
+
+// ResetToken packages a user ID with random data and an expiration time so that it can
+// be serialized and hashed into a token which can be sent to users.
+type ResetToken struct {
```
pdeziel, line 104: I was thinking of reusing the email verification fields in the database to store the secret, since I think we will not be sending reset password emails to unverified emails right?

pkg/quarterdeck/db/tokens.go
```diff
@@ -57,29 +52,76 @@ func (t *VerificationToken) Sign() (_ string, secret []byte, err error) {
 		return "", nil, err
 	}
 
-	// Compute hash with a random 64 byte key
-	key := make([]byte, keyLength)
-	if _, err = rand.Read(key); err != nil {
-		return "", nil, err
+	return t.signData(data)
```
pdamodaran, line 55: I like the consolidation into a separate function error

pkg/quarterdeck/db/tokens.go
```diff
@@ -57,29 +52,76 @@ func (t *VerificationToken) Sign() (_ string, secret []byte, err error) {
...
+
+	token = &ResetToken{
+		UserID: id,
+	}
+
+	if token.SigningInfo, err = NewSigningInfo(time.Minute * 15); err != nil {
+		return nil, err
+	}
+
+	if _, err = rand.Read(token.Nonce); err != nil {
```
pdamodaran, line 96: I am curious how this works because `Nonce` is not a direct attribute on `ResetToken`.  Also, is this check redundant since this check is being done within `NewSigningInfo`?

pdamodaran (approved):
Nice work @pdeziel!  I like how you consolidated some of the functions that are common to both the email verify token and the reset password token.  I had one comment but otherwise this is good to go.

pkg/quarterdeck/db/tokens.go
pdeziel: Nice catch, I didn't mean to be setting that field twice! The `Nonce` can be accessed like this if you use an embedded struct - note that `SigningInfo` is not specified with a name so `ResetToken` will automatically have all the fields and methods defined on `SingingInfo`. It's a useful pattern when you have very similar structs or structs that have similar functionality. Here's an example: https://gobyexample.com/struct-embedding

pdamodaran: ah - thanks for providing an example!

PR #783, opened 2023-10-14
Memory Efficient Task Scheduler and Rehash Task

bbengfort:
### Scope of changes

Implements a task to update event duplicates when policy changes. 

However, in order to implement this I needed to implement a task manager into Ensign. I reviewed the task manager in Quarterdeck and realized we had some memory efficiency problems with the scheduler.

I reimplemented the scheduler and renamed the `tasks` package to `radish` (e.g. radish not celery … and a nod to https://github.com/rotationalio/radish). 

Fixes SC-20996

### Type of change

- [x] new feature
- [ ] bug fix
- [ ] documentation
- [ ] testing
- [x] technical debt
- [ ] other (describe)

### Acceptance criteria

When the topic deduplication mechanism changes, we should create a background task that updates all of the hashes for the events in that topic.

This PR will likely be merged without review.

### Definition of Done

- [ ] ~I have recompiled and included new protocol buffers to reflect changes I made if necessary~
- [ ] ~Front-end: Checked sm, md, lg screen resolutions for effective responsiveness~

### Reviewer(s) checklist

- [ ] ~Front-end: I've reviewed the Figma design and confirmed that changes match the spec.~


pkg/utils/radish/scheduler.go
bbengfort, line 1: The old `tasks.Scheduler` had a comment "make sure there are no memory leaks" … so I checked, and there were. I created a sorted list of futures for handling future scheduled tasks and added a `Resize` method to ensure that the GC would clean up old tasks without increasing the capacity to the max number of tasks in the queue.

pkg/utils/radish/scheduler.go
```diff
@@ -0,0 +1,234 @@
...
+		s.Unlock()
+		return
+	}
+
+	s.running = true
+	s.Unlock()
+	s.run()
+}
+
+func (s *Scheduler) run() {
```
bbengfort, line 97: The nice thing about this scheduler is that it doesn't check every second if  tasks need to be run, it uses a timer to wait until the next task needs to be run and can be interrupted if a new task is added before that task.

PR #805, opened 2023-11-02
Uses calc to set a min height on the landing page's content

daniellemaxwell:
### Scope of changes

This change ensures that the footer sticks to the bottom of the screen on all screen sizes.

**Testing the Changes:**

To test the changes, you'll need to use `yarn`. If you haven't installed `yarn` before, run the following command in the web/beacon-app directory:
`yarn install`

Then run the following command to view the app locally:
`yarn dev`

Note: You will not need to run the containers to view these changes. The footer only appears before user's log in.

Fixes SC-22005

### Type of change

- [ ] new feature
- [x] bug fix
- [ ] documentation
- [ ] testing
- [ ] technical debt
- [ ] other (describe)

### Acceptance criteria

Note: I used the Sizzy browser to view the changes at different device sizes as I wasn't always able to see the issue while adjusting the screen in Chrome. Below, I've included screenshots for devices where I noticed the problem. The screen sizes are:
`1024 x 1366` and `2560 x 1440`

1024 x 1366

2560 x 1440

### Definition of Done

- [x] Front-end: Checked sm, md, lg screen resolutions for effective responsiveness

### Reviewer(s) checklist

### DIFF
--- a/web/beacon-app/src/components/auth/LandingFooter/LandingFooter.tsx
+++ b/web/beacon-app/src/components/auth/LandingFooter/LandingFooter.tsx
@@ -9,10 +9,10 @@ import TwitterIcon from '@/assets/icons/twitterIcon';
 function LandingFooter() {
   const { version: appVersion, revision: gitRevision } = appConfig;
   return (
-    <footer className="bg-footer bg-cover bg-no-repeat text-white ">
-      <div className="pt-72 2xl:pt-80">
+    <footer className="mt-20 bg-footer bg-cover bg-no-repeat text-white sm:mt-0 ">
+      <div className="pt-[300px] sm:pt-64 lg:pt-[225px] 2xl:pt-[320px]">
         <div className="mx-auto max-w-7xl">
-          <div className="mx-auto grid-cols-3 text-center sm:ml-0 sm:grid sm:text-left">
+          <div className="mx-auto grid text-center md:grid-cols-3">
             <div className="pt-4 font-bold leading-loose">
               <h3 className="font-light">PRODUCT</h3>
               <ul>
--- a/web/beacon-app/src/components/layout/MainLayout.tsx
+++ b/web/beacon-app/src/components/layout/MainLayout.tsx
@@ -8,7 +8,7 @@ const MainLayout = () => {
   return (
     <>
       <LandingHeader />
-      <Container className="place-items-center">
+      <Container id="landing-content" className="place-items-center">
         <Outlet />
       </Container>
       <LandingFooter />
--- a/web/beacon-app/src/index.css
+++ b/web/beacon-app/src/index.css
@@ -153,6 +153,10 @@
   }
 }
 
+#landing-content {
+  min-height: calc(100vh - 600px);
+}
+
 .RightSlot {
   margin-left: auto;
   padding-left: 20px;


web/beacon-app/src/components/auth/LandingFooter/LandingFooter.tsx
```diff
@@ -9,10 +9,10 @@ import TwitterIcon from '@/assets/icons/twitterIcon';
 function LandingFooter() {
   const { version: appVersion, revision: gitRevision } = appConfig;
   return (
-    <footer className="bg-footer bg-cover bg-no-repeat text-white ">
-      <div className="pt-72 2xl:pt-80">
+    <footer className="mt-20 bg-footer bg-cover bg-no-repeat text-white sm:mt-0 ">
```
daniellemaxwell, line 12: I noticed the footer was a little too close to the end of the content on smaller screens, so `mt-20` was added for a little breathing room. The style is removed once the `min-width` is 640px.

web/beacon-app/src/components/auth/LandingFooter/LandingFooter.tsx
```diff
@@ -9,10 +9,10 @@ import TwitterIcon from '@/assets/icons/twitterIcon';
...
+      <div className="pt-[300px] sm:pt-64 lg:pt-[225px] 2xl:pt-[320px]">
         <div className="mx-auto max-w-7xl">
-          <div className="mx-auto grid-cols-3 text-center sm:ml-0 sm:grid sm:text-left">
+          <div className="mx-auto grid text-center md:grid-cols-3">
```
daniellemaxwell, line 15: Adding `md` before `grid-cols-3` ensures that the style is only applied to screens with a `min-width` of 768px. The styles that were removed were no longer needed.

web/beacon-app/src/components/auth/LandingFooter/LandingFooter.tsx
```diff
@@ -9,10 +9,10 @@ import TwitterIcon from '@/assets/icons/twitterIcon';
 function LandingFooter() {
   const { version: appVersion, revision: gitRevision } = appConfig;
   return (
-    <footer className="bg-footer bg-cover bg-no-repeat text-white ">
-      <div className="pt-72 2xl:pt-80">
+    <footer className="mt-20 bg-footer bg-cover bg-no-repeat text-white sm:mt-0 ">
+      <div className="pt-[300px] sm:pt-64 lg:pt-[225px] 2xl:pt-[320px]">
```
daniellemaxwell, line 13: Without padding, the text below is positioned closer to the top of the footer element where the tower begins in the bg image. That area is full of whitespace, so the padding is pushing the `div` element containing the text down so that it appears above the blue background after the waves begin.

The `bg-cover` style was applied to make the bg image span the full width of a screen. Unfortunately, this causes the size of the waves to expand as the screen width increases. This then causes the text to sometimes appear above the waves, so that's why the padding changes depending on the screen size.

There may be a way to resolve the issue by cropping an image of the waves and then using it as a background to the bg image, but that will take much longer, so a follow on story will need to be created.

bbengfort: Ok, I understand -- let's keep it the way it is, worrying about ultra wide resolution is a pretty low priority. The fixes you've made so far have a higher impact. I don't think we need to make a follow on story for a cropped image of the waves.

bbengfort (approved):
I ran it locally and the changes that you made look great! Thank you so much for providing such detailed comments for the reviewer; it was very helpful!

PR #814, opened 2023-11-22
Prevents users from logging out before accepting invitations to join a team

daniellemaxwell:
### Scope of changes

Prevents users from having to log out to accept an invitation request while logged into Beacon.

Fixes SC-21664

Note: When the user accepts an invite, a new tab or window opens, and they're logged out of their previous session. I can add a follow on story that displays an error explaining why the user is logged out if they return to the previous tab or window.

### Type of change

- [x] new feature
- [ ] bug fix
- [ ] documentation
- [ ] testing
- [ ] technical debt
- [ ] other (describe)

### Acceptance criteria

https://www.awesomescreenshot.com/video/22679550?key=d116cb41d88cda4cfa85bfdc929ad5de

### Definition of Done

### Reviewer(s) checklist

### DIFF
--- a/web/beacon-app/src/features/teams/api/invitationAuthenticationRequest.ts
+++ b/web/beacon-app/src/features/teams/api/invitationAuthenticationRequest.ts
@@ -3,9 +3,9 @@ import { getValidApiResponse, Request } from '@/application/api/ApiService';
 import { APP_ROUTE } from '@/constants';
 import { UserAuthResponse } from '@/features/auth';
 
-export function getInviteAuthenticationRequest(
+export function invitationAuthenticationRequest(
   request: Request
-): ApiAdapters['getInviteAuthenticationRequest'] {
+): ApiAdapters['getInvitationAuthentication'] {
   return async (token: string) => {
     const response = (await request(`${APP_ROUTE.INVITE}/accept`, {
       method: 'POST',
--- a/web/beacon-app/src/features/teams/components/ExistingUserInvitationPage.tsx
+++ b/web/beacon-app/src/features/teams/components/ExistingUserInvitationPage.tsx
@@ -1,4 +1,4 @@
-import { Heading, Toast } from '@rotational/beacon-core';
+import { Heading } from '@rotational/beacon-core';
 import { useEffect } from 'react';
 import { useNavigate, useSearchParams } from 'react-router-dom';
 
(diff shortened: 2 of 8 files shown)


web/beacon-app/src/features/teams/components/ExistingUserInvitationPage.tsx
```diff
@@ -25,30 +28,47 @@ export default function ExistingUserInvitationPage({ data }: { data: any }) {
...
+      const token = decodeToken(login.auth.access_token) as any;
+      Store.setAuthUser(token, !!login.authenticated);
+
+      navigate(APP_ROUTE.DASHBOARD);
+    }
   }
+
   return (
     <div>
-      {login.hasAuthFailed && (
```
daniellemaxwell, line 45: This change removed a duplicate toast message.

web/beacon-app/src/features/teams/api/invitationAuthenticationRequest.ts
```diff
@@ -0,0 +1,19 @@
...
+): ApiAdapters['getInvitationAuthentication'] {
+  return async (token: string) => {
+    const response = (await request(`${APP_ROUTE.INVITE}/accept`, {
+      method: 'POST',
+      data: {
+        token,
+      },
+    })) as any;
+
+    return getValidApiResponse<UserAuthResponse>(response);
```
daniellemaxwell, line 17: The `UserAuthResponse` is listed [here](https://github.com/rotationalio/ensign/blob/develop/web/beacon-app/src/features/auth/types/LoginService.ts).

pdeziel (approved):
Thanks for fixing this - I think logging the user out of the previous tab is a reasonable side effect for now since that session would be expired anyway. It actually might be a good idea to add a generic "your session has expired, please login again" message to the login page, I'm not sure if we have that yet but that's what other sites do.

I attempted to locally reproduce the 500 error you were seeing but I had a problem where the verify email page was blank. I'm not sure if that's only on my end so it might be a good idea to try that again.

I did notice that `/verify` was being called twice as well...so there might be a larger issue we've discovered with the frontend issuing duplicate requests to the backend? If that's the case I'm satisfied with opening a bug story to look into that from the frontend perspective so we can get this merged in.

daniellemaxwell:
@pdeziel Were you directed to the `/verify` page instead of the `/invite` page after accepting an invite that was sent to an existing user? That shouldn't happen, so I want to be sure that I can reproduce the error.

